PR #234, opened 2015-01-16
remove dash default, allow any pattern length

ansis:
Necessary style spec changes for dashed line changes:
https://github.com/mapbox/mapbox-gl-js/pull/938
https://github.com/mapbox/mapbox-gl-native/pull/761


jfirebaugh:
Merged into #238.

PR #263, opened 2015-03-12
Use json-stringify-pretty-compact to format styles

### LINKED ISSUE #261: Format styles without js-beautify
js-beautify is a big and slow dependency. we should switch to https://github.com/lydell/json-stringify-pretty-compact

tristen:
- Closes #261

### DIFF
--- a/lib/format.js
+++ b/lib/format.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var beautify = require('js-beautify').js_beautify,
+var stringify = require('json-stringify-pretty-compact'),
     reference = require('../reference/v7.json'),
     sortObject = require('sort-object');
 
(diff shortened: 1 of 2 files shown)


lib/format.js
```diff
@@ -26,10 +26,7 @@ module.exports = function format(style) {
         return sortObject(layer, sameOrderAs(reference.layer));
     });
 
-    var str = beautify(JSON.stringify(style), {
-        indent_size: 2,
-        keep_array_indentation: true
-    }).replace(/"filter": {[^}]+}/g, function (str) {
+    var str = stringify(style).replace(/"filter": {[^}]+}/g, function (str) {
```
jfirebaugh: With `json-stringify-pretty-compact `, can we remove all of the custom string processing and just do `var str = stringify(style)`? What's the diff look like if you make that change and reformat mapbox-gl-styles?

jfirebaugh:
This ends up doing things like:

```
"paint": {"background-color": "@land"}
```

I think I like the existing formatting better, where most things are guaranteed to be on their own line, with exceptions for certain specific properties.

tmcw:
Can we just use JSON.stringify and if people really want to edit JSON then they just run it through js_beautify on their own? js_beautify is a big and slow dependency that only addresses a very rare and narrow usecase: I don't think it's worth the price.

jfirebaugh:
What's the price? This isn't something that goes into browserified files or something like that. Is it really that big a deal?

tmcw:
> What's the price? This isn't something that goes into browserified files or something like that. Is it really that big a deal?

It does exactly that: requiring this mapbox-gl-style-spec, which we do in app-styles, browserifies the dependency.

jfirebaugh:
Oh right, for export.

Let's just use plain ol `JSON.stringify` with a two-space indent then.

Issue #256, opened 2015-03-10
label layer priority order


ansis:
The list of layers in the style is bottom up. The last layer is drawn above the first layer.

Currently labels in earlier layers will block labels in later layers. This seems weird considering earlier layers are below later layers.

Should we make later layers block labels in earlier ones?

The order isn't specified in the spec but this styles would still need to be migrated.

ansis:
I talked with @nickidlugash and @peterqliu and they agree that we should reverse the order that labels are placed. This would mean that labels that are more important would come later in the style sheet. For example, country labels would come after road labels.

@jfirebaugh does adding this to v8 sound ok?

To migrate, all symbol layers would need to be reversed. This wouldn't work perfectly if a symbol layer is between two non-symbol layers, but I don't think we have any of those in real styles.

jfirebaugh:
Adding this to v8 sounds good to me.

peterqliu:
>This wouldn't work perfectly if a symbol layer is between two non-symbol layers, but I don't think we have any of those in real styles.

It's currently used in Emerald for [state labels](https://github.com/mapbox/mapbox-gl-styles/blob/mb-pages/styles/emerald-v7.json#L483-L512), but those are meant to be in the background and obstructed if necessary

jfirebaugh:
Removing from v8; not critical to shipping /app.

ansis:
-js implementation: https://github.com/mapbox/mapbox-gl-js/tree/v8-label-priority
migration script: https://github.com/mapbox/mapbox-gl-style-spec/tree/v8-label-priority
test-suite: https://github.com/mapbox/mapbox-gl-test-suite/tree/v8-label-priority

Migration script was tested against mapbox-gl-styles and everything looks good.

I think it would be good to get this for v8 because it's a breaking change. For each symbol layer we can either preserve it's priority related to other label layers or it's draw order. Not both. The migration script chooses the right thing to preserve in all the example styles, but there may be unintended effects. It might be good to make the change before we have more user styles to migrate.

Should I double check and merge tomorrow, or should this wait for after v8?

nickidlugash:
> This wouldn't work perfectly if a symbol layer is between two non-symbol layers, but I don't think we have any of those in real styles.

[...]

Issue #255, opened 2015-03-10
get rid of `text-max-size` and `icon-max-size`


ansis:
currently
`text-max-size` is a layout property and is used for label placement
`text-size` is a paint property is used to change

We should only have one property.

How would we deal with a property that is need both for layout and rendering?

jfirebaugh:
layout could calculate what the actual maximum `text-size` is going to be, right?

mourner:
Yeah, the only problem is that we introduce a new concept here: a paint property that implicitly affects the layout step as well. But it's probably fine.

jfirebaugh:
Hrm, yeah, `setPaintProperty` would then have to special-case `text-size` to trigger a relayout.

mourner:
Or we could have it the other way around — make it a layout property that implicitly creates a paint property.

jfirebaugh:
It would still need to support interpolated functions, which would be a different special case. But maybe a less problematic one.

nickidlugash:
@jfirebaugh @mourner @ansis any consensus on what we should do here?

nickidlugash:
I added this to [v8 omnibus](https://github.com/mapbox/mapbox-gl-style-spec/pull/264).

nickidlugash:
We should also get rid of `icon-max-size`. I've added that to this ticket name, and will add to v8.

jfirebaugh:
@ansis, can you look into this when you get back?

ansis:
pull request for -js: https://github.com/mapbox/mapbox-gl-js/pull/1419

Is the rest of v8 implemented in -native?

Issue #220, opened 2014-11-20
Treat "images" as nested collection of a style, like sources or layers


jfirebaugh:
We want something more convenient than the current automagical practice of appending permutations of `.json`, `.png`, and `@2x` to the `sprite` property and making requests to two separate resources which need to be in sync.

Options that have been proposed:
- **External protobuf container** A format inspired by glyph pbfs. The protobuf message would be a series of [metadata, image data] tuples. Metadata would be id, image format, pixel ratio, and sdf boolean. Open question: would we still have separate @1x and @2x files, or just use @2x images everywhere, or permit including multiple images per id (with different pixel ratios) in the same file?
- **Inline JSON + base64 image data** Include sprite data directly in the style. Metadata in JSON format, image data base64 encoded ([can't really do better than that](http://stackoverflow.com/questions/1443158/binary-data-in-json-string-something-better-than-base64)). Same question regarding @2x images as above.

@mapbox/gl

jfirebaugh:
I think we should:
- Eliminate the "sprite" concept and terminology. Instead a style has a collection of icons.
- Icons are embedded directly as an `"icons"` property of the style.
- An icon has a style spec JSON representation consisting of dimensions, pixel ratio, content type, and base64 encoded data. Content type must be PNG for now.
- The style API in gl-js (https://github.com/mapbox/mapbox-gl-js/issues/2058 / https://github.com/mapbox/mapbox-gl-js/issues/2059) and gl-native will be a standard CRUD API for the "icon" subresource of a style.
- `api.mapbox.com/styles/v2` renames the "sprite" subresource to "icons", but the interface is basically the same.

ansis:
> Icons are embedded directly as an "icons" property of the style.

Would there be different styles for different device pixel ratios?

jfirebaugh:
Multi-resolution support is TBD. But for starters, I think we could have Mapbox APIs always serve styles containing `@2x` icons, and for homesteaders, leave it up to them what resolutions they include with a style.

scothis:
> > Icons are embedded directly as an "icons" property of the style.
> 
> Would there be different styles for different device pixel ratios?

If we make inlining optional, we can preserve a link to the icons resource that can be loaded resolution independent, or with a new images format. There are advantages to inlining and externalizing depending on the circumstance.

ansis:
> I think we could have Mapbox APIs always serve styles containing @2x icons

[...]

PR #264, opened 2015-03-12
v8 Omnibus

### LINKED ISSUE #215: Rename *-image properties to *-pattern
Since `background-image`, `line-image` and `fill-image` tile the image, let's rename them to `background-pattern`, `line-pattern` and `fill-pattern`.

### LINKED ISSUE #247: centerpoint
Style objects should have a `center` key. This could be optional. If it were to follow the pattern used by the Mapbox Map API, would look like:

```
"center": [ 
  {longitude}, 
  {latitude},
  {center},
  {bearing}
]
```

### LINKED ISSUE #250: Switch to lon, lat order
* `center`
* video `coordinates`

### LINKED ISSUE #252: rename `symbol-min-distance` to `symbol-spacing`
I copied the name from carto but I copied it from the wrong property.

In carto `text-spacing` determines the distance between repeated labels on a line. `text-min-distance` does something else.

### LINKED ISSUE #257: Font stacks should be arrays
Font stacks shouldn't be strings concatenated by `,`s: this makes constants of fontstacks weird. Fontstacks should be string arrays.

### LINKED ISSUE #260: video source: url ⇢ urls
The `url` property of a video source is an array, it should be plural.

tmcw:
Patches:

[JS](https://github.com/mapbox/mapbox-gl-js/pull/1192), [Native](https://github.com/mapbox/mapbox-gl-native/pull/2052)

Changes

* [x] Font stacks as arrays. Fixes #257.
  * [x] JS https://github.com/mapbox/mapbox-gl-js/pull/1081
  * [x] Native https://github.com/mapbox/mapbox-gl-native/pull/993
  * [x] Spec https://github.com/mapbox/mapbox-gl-style-spec/blob/2b7c3777943049660613778e2076703c59a5aa9c/reference/v8.json#L462-L471
  * [x] Migration https://github.com/mapbox/mapbox-gl-style-spec/blob/2b7c3777943049660613778e2076703c59a5aa9c/migrations/v8.js#L44-L50
  * [x] Tests
  * [x] Styles
* [x] Permit map properties (centerpoint) Fixes #247 
  * [x] JS: no changes
  * [x] Native: no changes
  * [x] Migration: no changes
  * [x] Spec
  * [x] Styles
* [x] Rename symbol-min-distance to symbol-spacing. Fixes #252 
  * [x] JS
  * [x] Native https://github.com/mapbox/mapbox-gl-native/pull/993
  * [x] Spec https://github.com/mapbox/mapbox-gl-style-spec/blob/2b7c3777943049660613778e2076703c59a5aa9c/reference/v8.json#L315-L326
  * [x] Migration https://github.com/mapbox/mapbox-gl-style-spec/blob/2b7c3777943049660613778e2076703c59a5aa9c/migrations/v8.js#L51-L53
  * [x] Tests
  * [x] Styles
* [x] Rename video url to urls. Fixes #260 
  * [x] JS
  * [x] Native: no gl-native patch needed: video not supported
  * [x] Spec https://github.com/mapbox/mapbox-gl-style-spec/blob/2b7c3777943049660613778e2076703c59a5aa9c/reference/v8.json#L137
  * [x] Migration https://github.com/mapbox/mapbox-gl-style-spec/pull/264/files#diff-d94eefc8f1c24000ab9c6b1aa11ed555R34
  * [x] Tests
  * [x] Styles
* [x] ImageSource source type
  * [x] JS https://github.com/mapbox/mapbox-gl-js/pull/1174
  * [x] Native **shipping without** https://github.com/mapbox/mapbox-gl-native/issues/1350
  * [x] Spec
  * [x] Migration **none needed: new feature**
  * [x] Tests https://github.com/mapbox/mapbox-gl-test-suite/issues/27
  * [x] Styles
* [x] Flip coordinate order of video coordinates. Fixes #250
  * [x] JS
  * [x] Native: no gl-native patch needed: video not supported
  * [x] Spec
  * [x] Migration
  * [x] Tests
  * [x] Styles
* [x] background-image ⇢ background-pattern, line-image ⇢ line-pattern, fill-image ⇢ fill-pattern to fix #215
  * [x] JS
  * [x] Native https://github.com/mapbox/mapbox-gl-native/pull/993
  * [x] Spec
  * [x] Migration
  * [x] Tests
  * [x] Styles
* [x] Circle type
  * [x] JS https://github.com/mapbox/mapbox-gl-js/pull/1286
  * [x] Native https://github.com/mapbox/mapbox-gl-native/pull/2053
  * [x] Spec
  * [x] Tests
  * [x] Styles
* [x] Remove `text-max-size` and `icon-max-size`
  * [x] JS
  * [x] Native
  * [x] Spec https://github.com/mapbox/mapbox-gl-style-spec/issues/255
  * [x] Migration
  * [x] Tests
  * [x] Styles
* [x] Drop constants and color ops #308
  * [x] JS
  * [x] Native
  * [x] Spec
  * [x] Migration
  * [x] Tests
  * [x] Styles
* [x] Change `mapbox://` URL semantics #309
  * [x] JS
  * [x] Native
  * [x] Spec
  * [x] Migration
  * [x] Tests
  * [x] Styles
* [x] Flip label layer priority order #256
  * [x] JS
  * [x] Native
  * [x] Spec
  * [x] Migration
  * [x] Tests
  * [x] Styles


tmcw:
@lbud @quinnvd one thing I've always kind of wondered about the less-style operations (we used them directly in cartocss) was that

* lighten & darken
* saturate & desaturate
* fadeout & fadein

Are basically duals: saying `saturate: 10%` means `desaturate: -10%` and you get the picture. Especially given that we'll be wrapping these in a UI, I think separating the negative and positive versions of the transform will just make downstream implementations weirder.

Thoughts?

lbud:
@tmcw hm, true, and I see how if we use a slider UI centered at neutral this would make the implementation weird. But I also wonder if consistency from a Carto/LESS background is important.
cc @nickidlugash @samanpwbb @edenh — ?

nickidlugash:
@lbud Personally I'm in favor of prioritizing UI implementation over consistency with carto/LESS.

samanpwbb:
@nickidlugash @lbud regardless of how we end up doing this in the style JSON, the UI can do exactly what we want it to do. I was curious about this so I sketched it up:

https://github.com/mapbox/app-styles/issues/1352

Basically: show more fields and assume positive input values, or show less fields and assume both positive and negative input fields. Even if we include all the filters, we can still hide the 'negative' versions of the filters in the UI. I do believe if we implement a UI more like sketch 1, having both positive and negative options available would be better usability:

As a user, I know I want something to be desaturated. I'm going to look for "desaturate" first not for "negative saturate".

tmcw:
2015-03-25 at 2 19 pm

image

samanpwbb:
@tmcw I agree with you here, lets drop the inverse filters. It'll be simpler to implement in the UI.

nickidlugash:
I was thinking sliders, too.

lbud:
Ok, so we'll pick `saturate`, `lighten`, and which fade one makes more sense to be positive? (I'm instinctively inclined to say `fadein` should be positive, although I would guess that's used less frequently…)

samanpwbb:
@lbud how about just 'fade' - positive values increase fade, negative values decrease fade?

tmcw:
Another modification for comment:

* I think that we should support **embedded sprite metadata**: instead of pointing to a URL that contains sprite metadata, that data should be nested under the sprite key of the stylesheet.

jfirebaugh:
> embedded sprite metadata

What's the rationale? I'm tentatively :-1: because previous discussions have pointed towards an omnibus format for sprites that combines images and metadata, which would either duplicate or obsolete style-embedded-metadata.

tmcw:
What's the rationale?

* One fewer HTTP request to show each map
* Easier & faster to do better stylesheet validation - validate that references to the sprite are valid
* Simpler to implement in app-styles (okay, showing my hand with that one)

incanus:
Sprite metadata in the style feels bloated. 

- We already support two pixel ratios, and might do more. Even if we handle SVG sprites, we will always do rasters of some sort and multi-DPI just makes for a messy style section. 

- To clarify @jfirebaugh's point, we've talked (https://github.com/mapbox/mapbox-gl-style-spec/issues/220) about a binary sprite storage format to reduce branching and increase performance. Being able to seek a sprite's metadata out of the store without having to pre-parse the whole store from a text format would unblock runtime sprite changes (such as a set of POI icons changing during a label language swap) or sprite attribute changes (such as toggling an icon theme as part of a POI level-of-detail change or a class toggle which affects icon style, such as color change in a night style). 

- We will invariably keep the imagery itself out of the style (base64 encoding or something similar would be a waste), so why split the imagery and metadata? One less HTTP request doesn't feel like a level trade given the complexity of having imagery and metadata in two places. 

If sprites aren't valid in a style, just skip showing them? Easy way to point out a problem.

tmcw:
/cc @mapbox/gl 

Can we schedule a v8 sprint directly after the gl-native code is shipped? To ship this, we need many more hands on deck.

tmcw:
Additionally: should we fold #276 Circle render type into this issue? I'm about to sketch out an implementation in gl-js

lbud:
Grabbing types on native.

lbud:
@tmcw should [`validate.constants`](https://github.com/mapbox/mapbox-gl-style-spec/blob/v8/lib/validate/parsed.js#L67) also enforce that all constants have types?

tmcw:
y, i'll implement that.

Issue #294, opened 2015-06-17
don't enforce ascending domain for ordinal scales


jfirebaugh:
This validation makes sense only for non-ordinal scales.

jfirebaugh:
Fixed in #295.

PR #295, opened 2015-06-17
Improve scale validation

### LINKED ISSUE #292: validate scale type
In v7, style properties have a "function" specification that indicates whether they support either interpolated or piecewise-constant functions. This was never validated, mostly because the two function types have the same syntax (well, nearly -- interpolated functions support `"base"` and piecewise-constant functions don't, which could be validated).

Scale types are a little more involved. We should specify for each style property what scale types are supported:

* ordinal
* power with "normal" rounding (like v7 "interpolated")
* power with "floor" rounding (like v7 "piecewise-constant")

For most style properties this is obvious based on the type, but it's best to be explicit, and I know of at least one outlier: `line-dasharray`, which looks like it might support interpolation but doesn't.

Once this is implemented we should remove the old `"function"` property from the v8 spec.

lucaswoj:
Resolves #292

### DIFF
--- a/lib/validate/parsed.js
+++ b/lib/validate/parsed.js
@@ -338,9 +338,11 @@ module.exports = function(style, reference) {
 
         validate.object(key, val, specScale);
 
-        for (var i = 1; i < val.domain.length; i++) {
-            if (val.domain[i] < val.domain[i - 1]) {
-                return error(key + '.domain', val.domain, 'domain elements must be in ascending order');
+        if (specScale === reference['scale_power_none'] || specScale === reference['scale_power_floor']) {
+            for (var i = 1; i < val.domain.length; i++) {
+                if (val.domain[i] < val.domain[i - 1]) {
+                    return error(key + '.domain', val.domain, 'domain elements must be in ascending order');
+                }
             }
         }
 
--- a/lib/validate/parsed.js
+++ b/lib/validate/parsed.js
@@ -2,7 +2,6 @@
 
 var parseCSSColor = require('csscolorparser').parseCSSColor;
 var format = require('util').format;
-var _ = require('lodash');
 
 module.exports = function(style, reference) {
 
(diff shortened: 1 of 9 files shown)


lucaswoj:
cc @jfirebaugh

lucaswoj:
Resolves #294 too

jfirebaugh:
This is looking good.

Because it's pulled into gl-js, validation is size-conscious. I'd like to avoid the lodash dependency. Can you do something like:

```
    validate.scale = function(key, val, spec) {
        var scaleType = unbundle(val.type) || 'exponential';

        // Run basic validations
        var scaleSpec = _.extend({}, reference['scale_' + scaleType], reference['scale']);
        ...
    }
```

The oddity you see with `unbundle` comes from the fact that the style JSON might be parsed with `json-lint-lines-primitives`, which uses boxed Strings, Numbers, etc. in order to be able to decorate them with [line numbers for error reporting](https://github.com/mapbox/mapbox-gl-style-spec/blob/ce9da2ddd4f6e368153f92a406380db28b10f3ac/lib/validate/parsed.js#L19-21). So down here we have to unwrap the values with `unbundle` before using `===` comparison.

jfirebaugh:
Hah, you beat me to it.

lucaswoj:
@jfirebaugh @ansis 

This now includes

 - `exponential`, `categorical`, and `interval` scale types
 - validation to ensure only discrete output scales are used with discrete style properties

and some other misc bug fixes and refactoring

jfirebaugh:
:+1: Please rebase and merge.

lucaswoj:
(Still unsure if @mourner wrote a Github bot or is actually omniscient)

mourner:
@lucaswoj that's just https://waffle.io/ acting on my behalf. I don't know why the master robotic overmind chose me for channeling it's will. Maybe I was the last one to log in there.

PR #298, opened 2015-06-18
Unrename scales to functions

lucaswoj:
... and rename the function types in the reference JSON

cc @jfirebaugh @ansis 

https://mapbox.slack.com/archives/gl/p1434660255010066


lucaswoj:
@jfirebaugh this is also ready for :eyes:

jfirebaugh:
`s/continuious/continuous/g`

lucaswoj:
Ah! Too much copy paste.

reference/v8.json
```diff
@@ -826,32 +786,31 @@
       "mix"
     ]
   },
-  "function": {
+  "function-v7": {
```
jfirebaugh: Is it necessary to keep this?

lucaswoj: Depends. Do we still want the ability to validate v7 styles?

lib/validate/parsed.js
```diff
@@ -50,12 +50,12 @@ module.exports = function(style, reference) {
 
         // Functions
         if (reference.$version < 8 && spec.function && type === 'object') {
```
jfirebaugh:
```
if (spec.function && type === 'object') {
    return validate[reference.$version >= 8 ? 'function' : 'function-v7'](key, val, spec);
}
```

reference/v8.json
jfirebaugh: Yes, but when validating a v7 style the `reference` that will be passed in [here](https://github.com/mapbox/mapbox-gl-style-spec/blob/696fda69a24f59ef578db47558840404712b5895/lib/validate/parsed.js#L6) will come from `reference/v7.json`. [See here](https://github.com/mapbox/mapbox-gl-style-spec/blob/696fda69a24f59ef578db47558840404712b5895/lib/validate.js#L41).

lucaswoj: Of course! How silly of me.

lucaswoj:
@jfirebaugh :ship:?

jfirebaugh:
:ship:

PR #310, opened 2015-07-06
Change documentation generation command to 'docs' from 'doc'

friedbunny:
The readme already says to `npm run docs`, this updates the command in `package.json` to match that expectation and common practice. Committing to `v8` so it doesn't get clobbered on merge.

/cc @tmcw


tmcw:
:+1:

Issue #251, opened 2015-03-06
add "metadata" property at top level and within layers


jfirebaugh:
This would be a space to store arbitrary properties, either user-specified or for tools such as the editor. Metadata stored by tools should always use a key prefix such as `"mapbox:"` to avoid collisions with user data.

Once this is in the spec and existing tooling / styles are converted, update mapbox-gl-style-lint to prohibit unknown top-level and layer properties.

scothis:
:+1:

jfirebaugh:
Fixed in #314.

PR #314, opened 2015-07-10
Arbitrary metadata for stylesheets and layers

scothis:
It's useful for tools to be able to store additional metadata within a
stylesheet. Adding a 'metadata' object to the '$root' and 'layer' types.

While any value can be stored within the metadata object, it will not
affect the rendering. A bandwidth optimized, read-only stylesheet could
safely drop the property.

Property names within the metadata object should be prefixed to avoid
collisions between different tools and authors (ie 'mapbox:{propname}').

Issue: #251


scothis:
@tmcw can I get your :eyes: on this

tmcw:
:+1: can do in 30 and i'll have a PR to tradesies code review

tmcw:
:+1: looks good to me.

Issue #306, opened 2015-06-29
text-rotation should be interpolated, not piecewise-constant


tmcw:
Rotation should tween between zoom values.

jfirebaugh:
Making `text-rotate` an interpolated property would mean that we'd need to redo text placement on every zoom change. We currently redo placement on every rotation change and it's reasonably performant, so maybe this is ok. cc @ansis 

This would also make `text-rotate` the first interpolated layout property, which adds some complexity.

Is there a real-world use case for this?

ansis:
I'd rather not have it redo placement on zoom change. Labeling is designed so that labels do not jump around much while zooming. Redoing every time zoom changes would break that.

Since needing to smoothly change `text-rotate` seems really rare, I'd rather not make big changes to implement it.

We could eventually make it a dual paint/layout property similar to `text-size` (#255) so that the actual rendered text's rotation is interpolated but collision placement only happens at whole zoom levels.

Unless it's causing problems elsewhere, I'd like to leave it as is for now and maybe come back to it after #255.

jfirebaugh:
I'm assuming the use case here is "flashy but impractical rotate effect on zoom". Closing for now.

tmcw:
This is actually a bug because [the current style spec represents text-rotate as interpolated](https://github.com/mapbox/mapbox-gl-style-spec/blob/6719860a4010c57b846ac69597a4ccb8112ca64e/reference/v7.json#L545).

jfirebaugh:
Actually I think `interpolated` in the style spec was correct, and [my comment above](https://github.com/mapbox/mapbox-gl-style-spec/issues/306#issuecomment-116848370) was partially incorrect.

Numeric layout properties *can* support `interpolated` functions, and `text-rotate` is far from the only example. It's just that the functions are only ever evaluated at integer zoom levels.

jfirebaugh:
Next actions here would be to revert https://github.com/mapbox/mapbox-gl-style-spec/pull/319 and add some text to https://www.mapbox.com/mapbox-gl-style-spec/#function explaining how evaluation of layout property functions works.

tmcw:
@jfirebaugh for my usecase - which consists of "showing a chart in a ui that visualizes how things are interpolated", text will not be good enough: we need a programmatic, explicit hint that tells us whether something is interpolated or not.

jfirebaugh:
[...]

PR #319, opened 2015-07-22
Make it clear that text-rotate is not an interpolated property

tmcw:
Fixes https://github.com/mapbox/mapbox-gl-style-spec/issues/306


mourner:
:+1:

PR #324, opened 2015-08-05
Style diff

scothis:
Proposal for what style diffing could look like within mapbox-gl-style-spec. This code has worked well with mapbox-gl-js. Particularly interested in thoughts from @jfirebaugh @incanus and @lucaswoj


jfirebaugh:
Makes sense to me to have this code live here if we need to reuse it (are we at that point now?). It should live within `lib` though, like the other sources.

scothis:
I started with the code in `lib`, but moved it out to make it easier to require a specific version of the differ. This is the same pattern that is used for migrations and reference.

For example:
`require('mapbox-gl-style-spec/diff/v8');`

vs:
`require('mapbox-gl-style-spec/lib/diff/v8');`

jfirebaugh:
My hunch is we will want to take an approach more like that of validation, where version differences are [captured in inline branches](https://github.com/mapbox/mapbox-gl-style-spec/blob/24790118b768c03680d4b16820aa849b271fe824/lib/validate/parsed.js#L214), rather than having a completely separate implementation for each version.

scothis:
That would simplify the layout a good bit. The downside is that we'd only ever be able to diff stylesheet of the latest version.  If we're ok with that, I'll make the change.

lucaswoj:
The idea and the implementation look solid. 

I have a nagging fear that, in the long run, making GL style changes more efficient using diffs will be more painful than doing so using smart caching of the underlying resources.

scothis:
Incremental updates to a live style is just one use case. Another would be comparing the cumulative changes before publishing a new style. Another could be differential sync of styles across multiple collaborators. I'm sure there are other use cases I haven't thought of yet.

Minimizing the number of changes should always be more efficient, but I'll shed :joy: when setStyle is fast.

PR #326, opened 2015-08-05
Reverted function nomenclature

lucaswoj:
Re: https://github.com/mapbox/mapbox-gl-js/pull/1409#issuecomment-128158357

cc @jfirebaugh


jfirebaugh:
Please revert v8 too -- we're sticking with the existing function nomenclature for now. Looks like this will also requires some validator reverts.

lucaswoj:
@jfirebaugh :eyes:

jfirebaugh:
:+1:

PR #327, opened 2015-08-06
Removed duplicate circle layout properties

lucaswoj:
Previously, all style properties were present as both layout AND paint properties, with some inconsistencies between them. 

cc @jfirebaugh 

related to https://github.com/mapbox/mapbox-gl-test-suite/pull/36


jfirebaugh:
:+1:

Issue #308, opened 2015-06-30
Drop constants and color-ops from v8


jfirebaugh:

jfirebaugh:
Deep breath. Constants are gone from the editor. Do we want to drop them from the spec and renderers?

lucaswoj:
Just had a good chat with @tmcw, @nickidlugash, and others about this.

It sounds like there's a good case for removing constants **and color ops** from the v8 spec because

 - app doesn't support constants or color-ops
 - the current implementation of color-ops hasn't been tested in the wild ("unused code is broken by default") 
 - changing the spec for color-ops after they have been released would create headaches
 - removing / restoring color-ops from gl-js or gl-native only involves [~10 LOC](https://github.com/mapbox/mapbox-gl-js/blob/v8/js/style/style_declaration.js#L94)
 - we don't want to lose any information when importing / exporting styles from app (i.e. if color-ops were in the spec but not supported in app)

lucaswoj:
Ref https://github.com/mapbox/mapbox-gl-style-spec/issues/38

lucaswoj:
I think the right way forward is to remove both constants and color-ops from v8. We might want to work them into a future release.

lucaswoj:
We'll need to back out these changes:

> * color operations https://github.com/mapbox/mapbox-gl-style-spec/issues/38
>   * JS https://github.com/mapbox/mapbox-gl-js/pull/1158
>   * Native https://github.com/mapbox/mapbox-gl-native/pull/1311
>   * Spec https://github.com/mapbox/mapbox-gl-style-spec/pull/271

mb12:
@jfirebaugh What is the reason for this change? Does that mean we have to start inlining a  constant that is used in multiple places in the style.json?  For e.g. it is better to refer to a constant like  this via a meaningful name rather than inline it. 
```
    "@motorway_casing_width": {
      "base": 1.2,
      "stops": [[5, 0.4], [6, 0.6], [7, 1.5], [20, 22]]
    },
```

jfirebaugh:
@mb12 We are keeping the style spec tightly focused around the core features needed by the renderers and by our online style editor (launching publicly soon). Constants don't fall into either of those categories. If you are editing styles by hand and want to use constants, you can use a preprocessing step with a tool like https://github.com/screee/screess.

PR #330, opened 2015-08-11
Removed constants from v8 

lucaswoj:
cc @jfirebaugh @ansis 

Partially resolves #308

PR #333, opened 2015-08-11
Remove color ops from v8

lucaswoj:
cc @jfirebaugh 

ref #308

### DIFF
--- a/lib/validate/parsed.js
+++ b/lib/validate/parsed.js
@@ -346,34 +346,14 @@ module.exports = function(style, reference) {
 
     validate.color = function(key, val) {
         var type = typeof_(val);
-        if (type !== 'string' && type !== 'array') {
-            error(key, val, 'color expected, %s found', type);
-        } else if (type === 'string') {
-            if (reference.$version > 7) {
-                if (val[0] === '@' && constants[val]) val = constants[val].value;
-            } else {
-                if (val[0] === '@' && constants[val]) val = constants[val];
-            }
-            if (parseCSSColor(val) === null) {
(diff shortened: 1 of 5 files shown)


lucaswoj:
Sorry about the chaos between #330 and this PR. Everything's in ship shape and ready for :eyes: now.

lib/validate/parsed.js
```diff
@@ -345,35 +345,19 @@ module.exports = function(style, reference) {
...
-            if (parseCSSColor(val) === null) {
-                error(key, val, 'color expected, "%s" found', val);
-            }
-        } else {
-            validate.enum(key + '[0]', val[0], reference.color_operation);
-
-            if (isNaN(val[1])) {
-                error(key, val, 'number expected, "%s" found', val[1]);
-            }
+        if (reference.$version <= 7 && val[0] === '@') {
```
jfirebaugh: I believe constant values will have already been resolved by this point.

migrations/v8.js
```diff
@@ -1,5 +1,22 @@
 'use strict';
 
+var Reference = require('../reference/v8');
+var ColorOps = require('color-ops');
```
jfirebaugh: No need to support color-ops in the migration. These weren't available in v7 and there are no v8 styles in the wild that would need to be migrated.

lucaswoj: :+1:

lib/validate/parsed.js
lucaswoj: :+1:

PR #335, opened 2015-08-12
Added fontstack URL migration

### LINKED ISSUE #309: sematics for mapbox:// glyphs urls in v8
Specify owner in URL rather than trying to guess this based on an access token or the owner of a project using the style. Would allow explicitly using shared, public `mapbox` fonts in addition to private user fonts.

lucaswoj:
fixes #309

### DIFF
--- a/migrations/v8.js
+++ b/migrations/v8.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var Reference = require('../reference/v8');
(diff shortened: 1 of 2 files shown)


lucaswoj:
:eyes: @jfirebaugh

jfirebaugh:
:+1:

PR #336, opened 2015-08-13
v8 label priority migration

ansis:
adds migration for https://github.com/mapbox/mapbox-gl-style-spec/issues/256


ansis:
wrong target branch

PR #345, opened 2015-08-15
Future proof url.parse() for font stack v8 migrations

scothis:
In new versions of Node and browserify `URL.parse()` URL encodes
properties on the resulting object. This has the effect of breaking the
migration script.

We can decodeURI the path segments we need to compare in a manner that is
backwards and future compatible.

Tested with:
- browserify
- Node 0.10
- IO 3.0


jfirebaugh:
> `URL.parse()` URL encodes properties on the resulting object

WTF?! How does that make *any* sense?

PR #348, opened 2015-08-17
Simplify layout/paint property types

scothis:
Types were made more specific to add extra type safety around constants.
Without constants, we no longer need these more specific types. Fewer,
well-known types makes tooling support easier.

- '*-enum' -> 'enum'
- '*-array' -> 'array'
- 'opacity' -> 'number'
- 'field-template' -> 'string'

### DIFF
--- a/lib/validate/parsed.js
+++ b/lib/validate/parsed.js
@@ -380,26 +380,6 @@ module.exports = function(style, reference) {
     validate.string = typeValidator('string');
     validate.boolean = typeValidator('boolean');
 
-    // new types in v8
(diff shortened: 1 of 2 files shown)


jfirebaugh:
:+1: -- technically this is a breaking change but we meant to revert these for v8, and I don't think it will actually break anything downstream, so I'm inclined to be semver-lenient. 
https://github.com/mapbox/mapbox-gl-style-spec/blob/a8d26dc24c6d7bf82eadefbff26399cbc14b9b39/lib/validate/parsed.js#L383-L401 can also be simplified.

PR #350, opened 2015-08-21
Add sprites URL change to CHANGELOG.md

lucaswoj:
@jfirebaugh @tmcw


jfirebaugh:
From a user standpoint, the relevant change is the introduction of `mapbox://` URL support for sprites in v8. Let's move it down to the 8.0.0 section and characterize it as an addition, not a format change.

lucaswoj:
Got it. I wasn't sure where the v8.1.0 line was being drawn.

PR #356, opened 2015-09-04
Updates to symbol render and layout property descriptions

nickidlugash:
/cc @jfirebaugh @ansis to check technical accuracy of descriptions, and /cc @tmcw @samanpwbb to check that descriptions make sense in UI context.

### DIFF
--- a/reference/v8.json
+++ b/reference/v8.json
@@ -403,7 +403,7 @@
       "type": "boolean",
       "function": "piecewise-constant",
       "default": false,
-      "doc": "If true, other symbols may be visible even if they collide with the icon.",
+      "doc": "If true, other symbols can be visible even if they collide with the icon.",
       "requires": [
         "icon-image"
       ]
@@ -565,7 +565,7 @@
       "default": 0,
       "units": "em",
       "function": "interpolated",
-      "doc": "Text tracking value.",
+      "doc": "Text tracking amount.",
       "requires": [
         "text-field"
       ]
@@ -696,7 +696,7 @@
       "type": "boolean",
       "function": "piecewise-constant",
       "default": false,
-      "doc": "If true, other symbols may be visible even if they collide with the text.",
+      "doc": "If true, other symbols can be visible even if they collide with the text.",
       "requires": [
         "text-field"
       ]
--- a/reference/v8.json
+++ b/reference/v8.json
@@ -377,7 +377,7 @@
       "minimum": 1,
       "function": "interpolated",
       "units": "pixels",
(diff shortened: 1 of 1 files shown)


reference/v8.json
```diff
@@ -394,7 +394,7 @@
       "type": "boolean",
       "function": "piecewise-constant",
       "default": false,
-      "doc": "If true, the icon will be visible even if it collides with other icons and text.",
+      "doc": "If true, the icon will be visible even if it collides with other previously drawn symbols.",
```
jfirebaugh: This can be more explicit and say "collides with symbols from layers which appear later in the style."

nickidlugash: Since UI-wise the layer ordering is flipped, I'm trying to avoid descriptions that reference layers as being "later/earlier" or "above/below" in the style. Maybe we should add a description in the general Layers section on how symbol layer priority works?  (On the UI side, we can explain it in guides).

reference/v8.json
```diff
@@ -403,7 +403,7 @@
       "type": "boolean",
       "function": "piecewise-constant",
       "default": false,
-      "doc": "If true, the icon won't affect placement of other icons and text.",
+      "doc": "If true, other symbols may be visible even if they collide with the icon.",
```
tmcw: `can` instead of `may`

reference/v8.json
```diff
@@ -412,7 +412,7 @@
       "type": "boolean",
       "function": "piecewise-constant",
       "default": false,
-      "doc": "If true, the symbol will appear without its icon, in spaces where the icon would make it too large to fit.",
+      "doc": "If true, text will display without their corresponding icons when the icon collides with other symbols and the text does not.",
```
tmcw: text is singular, but `their` refers to a plural

reference/v8.json
```diff
@@ -493,7 +493,7 @@
         0
       ],
       "function": "interpolated",
-      "doc": "Icon's offset distance. Values are [x, y] where negatives indicate left and up, respectively.",
+      "doc": "Offset distance of icon from its anchor. Positive values indicate right and down, while negative values indicate left and up.",
```
tmcw: Why switch the word order?

> Icon's offset from its anchor.

Is clearer

reference/v8.json
```diff
@@ -565,7 +565,7 @@
       "default": 0,
       "units": "em",
       "function": "interpolated",
-      "doc": "Text kerning value.",
+      "doc": "Text tracking value.",
```
tmcw: could this be `amount`, or drop the "value"? "value" is redundant

PR #358, opened 2015-09-14
add explanation of tile size to pattern docs

samanpwbb:
Would be nice if docs explained to users the relationship between tiles and image dimensions. 

@nickidlugash @jfirebaugh let me know what you guys think of this language. I wasn't really sure how to talk about tile size. Can we say with more certainty that tile sizes are 512 rather than saying , "typically...".

### DIFF
--- a/reference/v8.json
+++ b/reference/v8.json
@@ -877,7 +877,7 @@
       "type": "string",
       "function": "piecewise-constant",
       "transition": true,
-      "doc": "Name of image in sprite to use for drawing image fills. For seamless patterns, image width and height must be a factor of two (2, 4, 8, 16, 32, and so on)."
+      "doc": "Name of image in sprite to use for drawing image fills. For seamless patterns, image width and height must be a factor of two (2, 4, 8, ..., 512)."
     }
   },
   "paint_line": {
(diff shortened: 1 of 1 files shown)


reference/v8.json
```diff
@@ -877,7 +877,7 @@
       "type": "string",
       "function": "piecewise-constant",
       "transition": true,
-      "doc": "Name of image in sprite to use for drawing image fills."
+      "doc": "Name of image in sprite to use for drawing image fills. For seamless patterns, image dimensions must be a factor of tile dimensions. Typically, tiles are 512px x 512px so image length and height should equal to 1, 2, 4, 8, 16, 32, 64, 128, 256, or 512."
```
jfirebaugh: "length and height" ⇢ "width and height"

jfirebaugh:
We can say with near certainty that tile sizes will always be a power of two >= 256. So 256, 512, or (someday) 1024.

samanpwbb:
Just tried to simplify the language, removed all mention of tile size, it felt confusing for an audience of nontechnical users who are interacting with a gl style via a UI, but I might have gone too far.

nickidlugash:
@samanpwbb I added the extended description for `line-pattern`, made the `background-pattern` description consistent with the rest, and added 512 as the high range for width/height (I think it's fine if technically the tiles could be 1024 or bigger, since even if they are bigger, it's doubtful that someone would need image files bigger than 512).

samanpwbb:
Looks great :+1:

PR #370, opened 2015-10-21
Make properties permalinkable.

### LINKED ISSUE #369: Make individual properties permalinkable

tmcw:
Fixes #369


jfirebaugh:
`visibility` is a corner case here, in that it's a property that all layers type have, and is named `visibility` in each. (We probably should have prefixed it like all the other properties.)

PR #382, opened 2015-11-26
Namespace navigation HTML IDs and anchors

buchanae:
As I've been learning Mapbox and referencing the style spec endlessly, I keep noticing that the nav doesn't link to the section I expect. For example, clicking "Sources" takes me to "Root.sources"

So I decided to fix that. Here ya go!

### DIFF
--- a/docs/_generate/index.html
+++ b/docs/_generate/index.html
@@ -7,60 +7,50 @@ class: fill-light
 options: full
 navigation:
   - title: Root
-    id: root
+    subnav:
+    - title: version
+    - title: name
+    - title: metadata
+    - title: center
+    - title: zoom
+    - title: bearing
+    - title: pitch
+    - title: sources
+    - title: sprite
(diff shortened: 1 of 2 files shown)


buchanae:
Oh, I also extended the navigation by adding the Root and Transition properties to sub navigation. And removed some code by using Liquid's `slugify` rather than defining an `id` for every nav item.

jfirebaugh:
Thanks! This looks like a good approach, but it doesn't quite seem to be working. The navigation sidebar generates anchors like `#Layers-Symbol`, while the href attribute uses `#layers-symbol`. When I click on the sidebar link, it doesn't jump to the right section (I guess anchors are case sensitive). Can you fix?

jfirebaugh:
Fixed and committed in 04d152b2dc551c0c5a7d822c9dc53135f4722180. Thanks!

PR #383, opened 2015-12-01
Add line-offset

kkaefer:
Offsetting lines is implemented in https://github.com/mapbox/mapbox-gl-native/pull/3092; we need to add it to the spec


lucaswoj:
This looks good overall. We might want to iterate on the doc a little bit to highlight 
- how `line-offset` differs from `line-translate`
- how `line-offset` is affected by the direction of the line

How about:

> The line's offset perpendicular to its direction. Values may be positive or negative, where positive indicates "leftwards" (if you were moving in the direction of the line) and negative indicates "rightwards."

lucaswoj:
ref https://github.com/mapbox/mapbox-gl-native/pull/3092

lucaswoj:
It looks like you're using this property to construct the line buffer. Should it be a layout property (instead of a paint property)?

kkaefer:
Where is it used  to construct a line Bucket?

lucaswoj:
Ah. I misunderstood the code. This is good as a paint property :+1:

kkaefer:
Updated the description.

lucaswoj:
Looks :shipit:

We need to figure out how to release this. Ideally we could release this as a minor release but this isn't _exactly_ backwards compatible because older versions of GL JS will fail validation on the unknown `line-offset` style property and throw an error. 

> - MINOR version when you add functionality in a backwards-compatible manner

I vote we release this as a minor release anyway and avoid using `line-offset` in any official styles until the next major release (token defaults?). Should we modify GL JS to not throw an error for an unknown style property? 

cc @jfirebaugh

lucaswoj:
per chat with @jfirebaugh, we are going to ship this as package version number `8.3.0`

kkaefer:
> Should we modify GL JS to not throw an error for an unknown style property?

Yeah, we should never fail, only print warning messages in this situation.

jfirebaugh:
I'm actually fine with throwing an error. I think it would be tough to tease the validations into a set that can safely be considered "warnings" versus those that have to be hard failures, and it's an edge case anyway. It's not that hard to use a renderer that supports the style properties you want to use.

PR #384, opened 2015-12-03
Fix nav links in docs

lyzidiamond:
This PR removes duplicate nav links in docs. cc @lucaswoj @jfirebaugh @danswick

### DIFF
--- a/docs/_generate/index.html
+++ b/docs/_generate/index.html
@@ -143,8 +143,8 @@ navigation:
     <% var source_types = ['vector', 'raster', 'geojson', 'image', 'video']; %>
 
     <div class='pad2 prose'>
-      <a id='sources-1' class='anchor'></a>
-      <h2><a href='#sources-1' title='link to sources'>Sources</a></h2>
+      <a id='root-sources' class='anchor'></a>
+      <h2><a href='#root-sources' title='link to sources'>Sources</a></h2>
       <p>
         Sources supply data to be shown on the map. The type of source is specified by the <code>"type"</code> property,
         and must be one of <var><%= source_types.join(', ') %></var>. Tiled sources (vector and raster) must specify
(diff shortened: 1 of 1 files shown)


danswick:
@lyzidiamond THANK YOU 

making those links so great

lucaswoj:
This is wonderful @lyzidiamond! Thank you :smile: 

I have an idea to make sure this doesn't happen again: nested the ids. For example, every subheading under "Sources" would have an id prefixed by `sources-`

```
- title: Root
    id: root
  - title: Sources
    id: sources
    subnav:
    - title: Vector
      id: sources-vector
    - title: Raster
      id: sources-rastersource
    - title: GeoJSON
      id: sources-geojson
    - title: Image
      id: sources-image
    - title: Video
      id: sources-video
```

It is a little more work up front (probably requiring adding a parameter to the [`item.html`](https://github.com/mapbox/mapbox-gl-style-spec/blob/headers/docs/_generate/item.html) template but it'll help us avoid making the same mistake in the future (or having to double check which one `sources-1` refers to).

lucaswoj:
It might be a little easier, and achieve the same affect, to just prefix all the top level ids with `root-`, like

```
- title: Root
    id: root-root
  - title: Sources
    id: root-sources
    subnav:
    - title: Vector
      id: vector
    - title: Raster
      id: rastersource
    - title: GeoJSON
      id: geojson
    - title: Image
      id: image
    - title: Video
      id: video
```

EDIT: ok, maybe `root-root` is unnecessary :stuck_out_tongue_winking_eye:

lucaswoj:
Feel free to :runner: with whichever you think is best.

lyzidiamond:
@lucaswoj i went with `root`! Ready to :ship:

docs/_generate/index.html
```diff
@@ -143,8 +143,8 @@
     <% var source_types = ['vector', 'raster', 'geojson', 'image', 'video']; %>
 
     <div class='pad2 prose'>
-      <a id='sources' class='anchor'></a>
-      <h2><a href='#sources' title='link to sources'>Sources</a></h2>
+      <a id='sources-1' class='anchor'></a>
```
lucaswoj: Do you need to change this to `<a id='root-sources' class='anchor'></a>` too?

lyzidiamond: lol yes

lyzidiamond:
okay @lucaswoj actually fixed now.

jfirebaugh:
Can we sync this with the approach in #382 and get a single working PR?

#382 has the virtue of auto-generating the IDs with `slugify`, but needs to use liquid's `downcase` filter so that cases match up.

lyzidiamond:
@jfirebaugh fine by me, just wanted to have it fixed! I'll close here in favor of #382.

PR #386, opened 2015-12-07
Line-offset behavior is opposite other offsets

kkaefer:
Line offset:

> positive indicates “leftwards” [...] and negative indicates “rightwards.”

Text and icon offset:

> Positive values indicate right and down, while negative values indicate left and up.

When using line-offset and adding icons to the line for labeling the direction of travel the icon-offset must be the inverse of the line-offset.

Shouldn't they be consistent?

cc @kkaefer @lucaswoj

### DIFF
--- a/reference/v8.json
+++ b/reference/v8.json
@@ -986,7 +986,7 @@
     "line-offset": {
       "type": "number",
       "default": 0,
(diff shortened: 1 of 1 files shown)


kkaefer:
Should we inverse the meaning? If so, we should do it right away.

jfirebaugh:
Yeah, seems like this should be flipped. Can we get this into new releases before folks start using the properties with their current meaning?

kkaefer:
Master ticket: https://github.com/mapbox/mapbox-gl/issues/3#issuecomment-162607343

PR #393, opened 2015-12-17
Added further clarification GeoJSON source docs

### LINKED ISSUE #1840: simplification tolerance option for layer source
https://www.mapbox.com/mapbox-gl-js/api/#GeoJSONSource
What does options.tolerance do?

Can these options be used together with map.addSource as well?

lucaswoj:
cc @mourner @musicformellons 

fixes https://github.com/mapbox/mapbox-gl-js/issues/1840

### DIFF
--- a/reference/v8.json
+++ b/reference/v8.json
@@ -158,17 +158,17 @@
     "maxzoom": {
       "type": "number",
       "default": 14,
-      "doc": "Maximum zoom to preserve detail at."
+      "doc": "Maximum zoom level at which to create vector tiles (higher means greater detail at high zoom levels)."
     },
     "buffer": {
       "type": "number",
       "default": 64,
-      "doc": "Tile buffer on each side."
+      "doc": "Tile buffer size on each side (higher means fewer rendering artifacts near tile edges but slower performance)."
     },
     "tolerance": {
       "type": "number",
       "default": 3,
-      "doc": "Simplification tolerance (higher means simpler)."
+      "doc": "Douglas-Peucker simplification tolerance (higher means simpler geometries and faster performance)."
     }
   },
   "source_video": {


reference/v8.json
```diff
@@ -158,17 +158,17 @@
     "maxzoom": {
       "type": "number",
       "default": 14,
-      "doc": "Maximum zoom to preserve detail at."
+      "doc": "Maximum zoom level at which to create vector tiles (higher means greater detail at high zoom levels but slower performance)."
```
mourner: Actually I'm not even sure it affects performance in a noticeable way. We should probably bump the default to 16 or even 18.

reference/v8.json
```diff
@@ -158,17 +158,17 @@
...
     },
     "buffer": {
       "type": "number",
       "default": 64,
-      "doc": "Tile buffer on each side."
+      "doc": "Tile buffer size (higher means fewer rendering artifacts near tile edges but slower performance)."
```
mourner: "on each side" may be an important clarification since it could also be read as "total buffer in one dimension".

reference/v8.json
lucaswoj: I myself have a pretty vague understanding of the implications of increasing the maxzoom of a GeoJSON tile source. Maybe you could take a swing at writing this doc property?

mourner: Eh, I'm not even sure they are significant enough to be written out. Let's just drop "but slower performance" part here (or everything inside the parentheses) and merge.

lucaswoj: :rocket:

lucaswoj:
Merged as 36afc71038509f700ba2c44c0cec5a34add9a805

PR #395, opened 2015-12-18
require ids on layers, closes #394

### LINKED ISSUE #394: layers with no id are valid?
Seems like a layer with no id should be caught by validator. Looking here: https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/lib/validate/parsed.js#L131

samanpwbb:
cc: @jfirebaugh


jfirebaugh:
Can you fix this by adding `"required": true` [here](https://github.com/mapbox/mapbox-gl-style-spec/blob/a1f0723014eb18fc182f275918f86f73b5a840ea/reference/v8.json#L232) instead?

Surprising that we never had this.

samanpwbb:
@jfirebaugh updated. I also bumped the version of this test fixture, if that makes sense.

jfirebaugh:
:+1:

PR #397, opened 2015-12-22
prevent ramp from having duplicate stops

mayagao:
@jfirebaugh Do I also need to create a test for this? Doesn't seem to have a test for ramps.

cc @samanpwbb

### DIFF
--- a/lib/validate/parsed.js
+++ b/lib/validate/parsed.js
@@ -328,8 +328,8 @@ module.exports = function(style, reference) {
                             error(key + '[0]', val[0], 'zoom level for piecewise-constant functions must be an integer');
                         }
 
-                        if (val[0] < lastStop) {
-                            error(key + '[0]', val[0], 'array stops must appear in ascending order');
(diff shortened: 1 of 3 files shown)


jfirebaugh:
Yes, insert a test case [here](https://github.com/mapbox/mapbox-gl-style-spec/blob/1b8a95179efc879712afe999dabc9b3ac03cb75a/test/fixture/functions.input.json#L127-L128), and then run `UPDATE=1 npm test`, which should update `functions.output.json` with the expected output.

mayagao:
is this the right way?

jfirebaugh:
The line number updates in `functions.output.json` are expected, but I'd also expect to see a new error with the text "array has duplicate stops" in the output. Since it's not there, for some reason the test case does not trigger the new validation. Can you try using `==` instead of `===`? The validator uses boxed types so that it can attach line numbers to them, which means that `===` comparison doesn't always work.

mayagao:
ok updated, thank you @jfirebaugh

PR #407, opened 2016-01-29
Refactor to modularize validator

lucaswoj:
I have rewritten the validator as a series of independent modules without global state to enable us to
- validate style API inputs in https://github.com/mapbox/mapbox-gl-js/issues/1632
- write better validator tests
- make our codebase more maintainable

The only planned remaining task is a through read-through / documentation revamp. As a follow up, we should think about refactoring our tests to exercise the independent modules instead of the whole system. 

This is, admittedly, _huge_ change in terms of LOC but I feel pretty good about our test coverage. 

cc @jfirebaugh @mourner @ansis


test/fixture/layers.output.json
```diff
@@ -52,10 +52,6 @@
     "line": 64
   },
   {
-    "message": "layers[11]: duplicate layer id \"duplicate\", previously used at line 68",
-    "line": 74
-  },
-  {
```
lucaswoj, line 58: This shouldn't be here.

ansis:
I haven't reviewed this carefully but it does look a lot less scary than what we had before!

lib/validate.js
```diff
@@ -1,75 +0,0 @@
-'use strict';
-
-var jsonlint = require('jsonlint-lines-primitives');
-var spec = require('../');
-var validate = require('./validate/parsed');
-
-/**
```
jfirebaugh, line 7: Looks like all the documentation in this file got lost in the shuffle.

lib/validate_style.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+var validateConstants = require('./validate/validate_constants');
+var validate = require('./validate/validate');
+var latestStyleSpec = require('../reference/latest.min');
+var parseStyle = require('./parse_style');
```
jfirebaugh: The previous file and export structure was carefully chosen and I'd prefer not to change it.

In particular, `validate_style.js` needs to _not_ `require('./parse_style')`, because that pulls in the entirety of the module and it's dependencies, which will bloat mapbox-gl-js.

See https://github.com/cutting-room-floor/mapbox-gl-style-lint/issues/48.

jfirebaugh: Also stuff like `require('mapbox-gl-style-spec/lib/validate/latest')` is part of the public API, and although semver in this module is based on the _style spec_ rather than the module API, I'd like to also avoid backwards incompatible changes to the module API outside of major style spec revisions if possible.

lucaswoj: Ah. I was thinking we included `jsonlint` in mapbox-gl-js (though, it seems pretty obvious that we wouldn't in retrospect). I will fix the situation.

lucaswoj: It sucks that our file structure is a "public API" :disappointed: 

Is this an acceptable compromise, @jfirebaugh?
- add a shim for `mapbox-gl-style-spec/lib/validate/latest` https://github.com/mapbox/mapbox-gl-style-spec/commit/405a3389d80464c0b5f4b6e009841d5c2217b51c
- add an explicitly minified style validation entrypoint `mapbox-gl-style-spec/lib/validate_style.min` https://github.com/mapbox/mapbox-gl-style-spec/commit/d059087beb1f3cfcc1c38841d1e241457473446b

If so, I will
1. documentation.js the _heck_ out of this module
2. build a solution for https://github.com/mapbox/mapbox-gl-js/issues/1632
3. merge this into master

lucaswoj:
Added docs to `validate_style` and `validate_style.min`. It feels like a distraction to document the granular validators right now. Lets punt on that until it becomes a pain point. 

Is there anything else you'd like to see here @jfirebaugh @ansis?

jfirebaugh:
Some validation functions return arrays of errors, some return a scalar error, and some return one or the other depending on internal code flow. Can we make them all return arrays consistently?

lucaswoj:
> Some validation functions return arrays of errors, some return a scalar error, and some return one or the other depending on internal code flow. Can we make them all return arrays consistently?

@jfirebaugh done in 21cebfd836e54b460bf4f4aecedd2a4e24c804e1

jfirebaugh:
:ship:

PR #411, opened 2016-02-05
Migrate ramps to ensure validity

scothis:
We added validation checks to make sure that ramp stops are sorted by
zoom level and do not have duplicate stops by zoom. This broke some
styles that have ramps with out of order and/or duplicate stops.

We should migrate these styles ramp to ensure compliance with the new
validations. The migration does the following:
- sorts ramp stops by zoom
- removes duplicate stops
- marginally increments the zoom of a colliding stop with a unique value
  - rerun the migration in case changing the zoom causes a new
    collision/ordering issue

### DIFF
--- a/migrations/v8.js
+++ b/migrations/v8.js
@@ -2,6 +2,7 @@
 
 var Reference = require('../reference/v8');
 var URL = require('url');
+var isEqual = require('lodash.isequal');
 
 function getPropertyReference(propertyName) {
     for (var i = 0; i < Reference.layout.length; i++) {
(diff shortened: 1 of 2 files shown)


scothis:
@tmcw for review

jfirebaugh:
When/where will this migration get run?

scothis:
> When/where will this migration get run?

Good question, typically we migrate between major version. The issue here is that we made a backwards incompatible change (https://github.com/mapbox/mapbox-gl-style-spec/pull/397) between 8.3 and 8.4, which causes some styles to not render. There's overhead in migrating a v8 style to v8, but a little overhead is better than a style breaking when upgrading to a new version of gl-js.

Minimally, we can run the migrator within gl-js if style validation errors are detected. Other renders will need to figure out their own migration path.

jfirebaugh:
Running a v8 ⇢ v8 migration is unfortunately unsafe -- for example it will incorrectly reverse coordinates in video sources and the order of symbol layers.

I think it might be prudent to revert #397 and let it wait for v9.

scothis:
> I think it might be prudent to revert #397 and let it wait for v9.

Agreed.

scothis:
Continued at #413

PR #413, opened 2016-02-05
Revert "prevent ramp from having duplicate stops"

scothis:
This reverts commit e37aa5f678eb472c47bc2388bcc8ec71330e1c3e.

Adding validation of ramp stops was good in spirit, but it breaks some v8
styles. Reverting until v9 where we can make backwards incompatible
changes.


jfirebaugh:
:+1:

PR #423, opened 2016-03-07
clarified icon-offset units are in pixels

charlieg:


jfirebaugh:
Thanks for the contribution! However, they're not pixels, exactly; see #84.

PR #424, opened 2016-03-07
Validate the presence and contents of glyphs and sprite urls

### LINKED ISSUE #170: Are sprites, glyphs, and constants required?
2014-09-22 at 11 42 am

### LINKED ISSUE #299: validate presence of glyphs/sprite when text/icons are used
Validation should catch an issue like https://github.com/mapbox/mapbox-gl-js/issues/1291.

lucaswoj:
fixes #299
fixes #170 

cc @jfirebaugh @ansis @bhousel @tmcw @mourner

### DIFF
--- a/lib/validate/validate_glyphs_url.js
+++ b/lib/validate/validate_glyphs_url.js
@@ -1,0 +1,22 @@
+'use strict';
+
+var ValidationError = require('../error/validation_error');
+var validateString = require('./validate_string');
+
+module.exports = function(options) {
+    var value = options.value;
+    var key = options.key;
+
+    var errors = validateString(options);
+    if (errors.length) return errors;
+
+    if (value.indexOf('{fontstack}') === -1) {
+        errors.push(new ValidationError(key, value, '"glyphs" url must include a "{fontstack}" token'));
+    }
+
+    if (value.indexOf('{range}') === -1) {
+        errors.push(new ValidationError(key, value, '"glyphs" url must include a "{range}" token'));
+    }
+
+    return errors;
+};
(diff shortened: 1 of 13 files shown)


test/fixture/missing-glyphs.output.json
```diff
@@ -0,0 +1,6 @@
+[
+  {
+    "message": "layers[0].layout.text-field: \"glyphs\" url must be specified",
```
jfirebaugh: Can this be a little more descriptive?

> the use of "text-field" requires a style with a "glyphs" property

lib/validate/validate_glyphs_url.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+var ValidationError = require('../error/validation_error');
+
+module.exports = function(options) {
```
tmcw, line 6: if glyphs is set to `true`, will this still run, or will that be caught ahead of time as a wrong type?

test/fixture/missing-glyphs.output.json
lucaswoj: :+1: changed to 

> use of "icon-image" requires a stylesheet "sprite" property
> use of "text-field" requires a stylesheet "glyphs" property

lib/validate/validate_glyphs_url.js
lucaswoj, line 6: Good catch. I added a call to `validateString` in this method.

test/fixture/missing-glyphs.output.json
jfirebaugh: `s/stylesheet/style`

lucaswoj: Oops. That makes sense. My apologies.

PR #443, opened 2016-04-16
update validation for new data + zoom driven functions

ansis:
:eyes: @lucaswoj

### DIFF
--- a/lib/validate/validate_function.js
+++ b/lib/validate/validate_function.js
@@ -5,9 +5,12 @@ var getType = require('../util/get_type');
 var validate = require('./validate');
 var validateObject = require('./validate_object');
 var validateArray = require('./validate_array');
+var validateNumber = require('./validate_number');
 
 module.exports = function validateFunction(options) {
     var originalValueSpec = options.valueSpec;
+
+    var stopKeyType;
 
     return validateObject({
         key: options.key,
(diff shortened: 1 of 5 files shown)


lucaswoj:
Per chat with @scothis, we're going to relax some of these restrictions. Ping me when this PR is ready for 👀.

ansis:
@lucaswoj this should be ready for :eyes:. 7bc1c53 relaxes restrictions. Do any changes need to be made the spec itself?

lib/validate/validate_function.js
```diff
@@ -51,13 +53,29 @@ module.exports = function validateFunction(options) {
...
+        }
+
+        if (type === 'object') {
+            errors = errors.concat(validate({
+                key: key + '[0].zoom',
+                value: value[0].zoom,
+                valueSpec: {type: 'number'},
+                style: options.style,
+                styleSpec: options.styleSpec
+            }));
```
lucaswoj, line 77: Should we also validate the `value` property, if it is present?

lucaswoj:
> Do any changes need to be made the spec itself?

Nope, you've captured it all in code 👍.

lib/validate/validate_function.js
lucaswoj, line 77: I think we can assert that
- `value` is either a string or a number
- no keys other than `value` and `zoom` are present

lucaswoj:
Was the change to `bin/gl-style-composite` an accident?

lucaswoj:
This is good to 🚢 once you add a test for the new zoom-and-feature function syntax.

ansis:
d45b8734c49dd8df446c94c1bfda3f1320830c9f adds validation tests for zoom+feature functions

PR #445, opened 2016-04-25
Allow string-keyed object stops for property functions

lucaswoj:
This function should not cause a validation error

``` js
{
  property: 'mapbox',
  stops: [['great', 1], ['swell', 2]]
}
```

Just as this function does not

``` js
{
  property: 'mapbox',
  stops: [[{value: 'great', zoom: 0}, 1], [{value: 'swell', zoom: 0}, 2]]
}
```

cc @ansis

### DIFF
--- a/lib/validate/validate_function.js
+++ b/lib/validate/validate_function.js
@@ -9,6 +9,7 @@ var validateNumber = require('./validate_number');
 
 module.exports = function validateFunction(options) {
     var originalValueSpec = options.valueSpec;
+    var originalValue = options.value;
 
     var stopKeyType;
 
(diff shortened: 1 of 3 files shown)


lib/validate/validate_function.js
```diff
@@ -76,10 +77,11 @@ module.exports = function validateFunction(options) {
                 objectElementValidators: { zoom: validateNumber, value: validateValue }
             }));
         } else {
-            errors = errors.concat(validate({
+            var isZoomFunction = originalValue.property === '$zoom' || !originalValue.property;
```
ansis: I think this should be just `var isZoomFunction = !originalValue.property;`

`$zoom` isn't supported by mapbox-gl-function

ansis:
Other than the `$zoom` detail this looks ready to be merged.

lucaswoj:
Ack! Yes! Sorry I keep tripping over that. 🚢

Issue #410, opened 2016-02-03
Throws if a layer references a source but the stylesheet doesn't include a sources property at all


tmcw:
``` json
{
    "version": 8,
    "name": "Basic",
    "sprite": "mapbox://sprites/tmcw/foo",
    "glyphs": "mapbox://fonts/mapbox/{fontstack}/{range}.pbf",
    "layers": [
        {
            "id": "landuse_park",
            "type": "fill",
            "source": "mapbox",
            "source-layer": "landuse",
            "filter": [],
            "paint": {
                "fill-color": "#d2edae"
            },
            "interactive": true
        }
    ],
    "id": "cik1j5my8000l7pkpig50p4xr"
}
```

This input triggers an exception [right here](https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/lib/validate/parsed.js#L163) because we try to look for `style.sources[val.source]` but `style.sources` is `undefined`.

@lucaswoj should this get fixed before or after or as part of #407 ?

lucaswoj:
Fixing this after #407 would be my ideal workflow. I hope #407 will land within the next couple days. If you want to submit a PR before #407 lands, I'm happy to backport the change.

tmcw:
@lucaswoj #407 has landed. Added a failing testcase for you in https://github.com/mapbox/mapbox-gl-style-spec/pull/447

PR #447, opened 2016-05-03
Failing testcase for #410

tmcw:


lucaswoj:
Thanks for the test case @tmcw. How does the fix look?

tmcw:
👍 lgtm!

PR #450, opened 2016-05-05
Update function documentation and validation for property functions

lucaswoj:
It's not perfect but it is a step in the right direction! 

cc @tmcw @lyzidiamond @danswick @jfirebaugh 

screen shot 2016-05-05 at 11 27 17 am
screen shot 2016-05-05 at 11 27 07 am

### DIFF
--- a/docs/_generate/index.html
+++ b/docs/_generate/index.html
@@ -550,48 +550,105 @@ navigation:
         <div class='pad2 keyline-bottom'>
           <a id='types-function' class='anchor'></a>
           <h3 class='space-bottom1'><a href='#types-function' title='link to function'>Function</a></h3>
-          <p>The value for any layout or paint property may be specified as a <em>function</em>, allowing you to
-              parameterize the value according to the current zoom level. Functions are written as an array of stops,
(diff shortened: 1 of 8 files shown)


docs/_generate/index.html
```diff
@@ -550,48 +550,105 @@ <h3 class='space-bottom1'><a href='#types-array' title='link to array'>Array</a>
...
+      // "temperature" is 100 -> circle color will be red
+      [100, 'red']
+
+    ]
+  }
+}
+              {% endhighlight %}
+          </div>
+
+          <p class='small'><strong>Zoom-and-feature functions</strong> allow the appearance of a map feature to change with both its properties <em>and</em> zoom. Each stop is an array with two elements, the first is an object with a property input value and a zoom, and the second is a function output value. Note that support for property functions is not yet complete.</p>
```
jfirebaugh: "Zoom-and-feature" ⇢ "Zoom-and-property"

tmcw:
> Note that support for property functions is not yet complete.

Can we be way more specific about how it's not complete?

PR #453, opened 2016-05-10
Make fill opacity language clearer. Fixes #216

### LINKED ISSUE #216: Unify color & opacity
This has always been kind of a puzzling thing: why do we have both fill-color and fill-opacity? Could we represent opacity with rgba always?

tmcw:
cc @jfirebaugh for the review


jfirebaugh:
👍

PR #457, opened 2016-05-30
change function type "continuous" to "exponential"

fuzhenn:
It seems to be "exponential" according to project https://github.com/mapbox/mapbox-gl-function


jfirebaugh:
cc @lucaswoj

lucaswoj:
Thanks for the fix!

PR #464, opened 2016-06-21
Docs for EPSG:3857 WMS support in Mapbox GL JS

bhousel:
Added a third bullet point about tiled WMS sources..

screenshot 2016-06-21 13 19 25


1ec5:
096c534fc6076b630c6469f8a1a9290e0dd10725 points out that the Mapbox macOS SDK also supports WMS URLs. The other SDKs will support them in their upcoming releases: mapbox/mapbox-gl-native#5628.

PR #465, opened 2016-07-01
Annotate style reference and documentation with SDK requirements

jfirebaugh:
Refs https://github.com/mapbox/mapbox-gl/issues/16.

Each layer type and property is annotated with the first version of the JS, iOS, and Android SDK that supports the feature. For properties, support is divided into basic support and data-driven styling support. Absence of an annotation for a particular SDK implies that the feature is unsupported. Unsupported combinations are marked as such and in some cases linked to the relevant GitHub issue.

Most features are annotated as supported starting with GL JS 0.10.0, iOS SDK 2.0.0, and Android SDK 2.0.1 -- the first stable releases that supported GL rendering of the initial v8 style specification. The exceptions are the following:

Source support:
- Image and Video sources
- GeoJSON clustering

Properties added to v8 after the initial release:
- line-offset
- text-pitch-alignment
- icon-text-fit
- icon-text-fit-padding

Properties where data-driven styling support has been introduced in GL JS:
- circle-opacity
- circle-blur
- fill-color
- fill-outline-color
- circle-color
- circle-radius
## TODO
- [ ] Filter operators are not yet annotated. The interesting operator is `has`/`!has`, which was a recent introduction.
- [ ] Did I miss anything?
- [ ] Needs a design pass.
## Screenshots

Example source:

image

Example property:

image

cc @mapbox/gl

### DIFF
--- a/docs/_generate/item.html
+++ b/docs/_generate/item.html
@@ -1,11 +1,11 @@
-<div class='col12 clearfix pad0y'>
+<div class='col12 clearfix pad0y pad2x'>
   <div>
     <span class='code space-right'><a id='<%= id %>' href='#<%= id %>'><%= name %></a></span>
     <% if (prop.function == "interpolated" ) { %><span class='icon smooth-ramp inline quiet'></span><% } %>
     <% if (prop.function == "piecewise-constant" ) { %><span class='icon step-ramp inline quiet'></span><% } %>
     <% if (prop.transition) { %><span class='icon opacity inline quiet'></span><% } %>
   </div>
-  <div class='small pad2x'>
+  <div>
     <em class='inline quiet'>
       <%= prop.required ? 'Required' : 'Optional' %>
       <% if (prop.type && prop.type !== '*') { %><a href='#<%= prop.type %>'><%= prop.type %></a>.<% } %>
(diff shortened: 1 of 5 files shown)


lucaswoj:
This looks great! Thank you for taking on this project @jfirebaugh. 🚢 once you've had a chance to iterate on the design.

docs/_generate/item.html
```diff
@@ -49,4 +49,26 @@
...
+      <td><%- prop['sdk-support'].basic.js ? '>= ' + prop['sdk-support'].basic.js : 'Not supported' %></td>
+      <td><%- prop['sdk-support'].basic.ios ? '>= ' + prop['sdk-support'].basic.ios : 'Not supported' %></td>
+      <td><%- prop['sdk-support'].basic.android ? '>= ' + prop['sdk-support'].basic.android : 'Not supported' %></td>
+    </tr>
+    <tr>
+      <td>Data-driven styling</td>
+      <td><%- (prop['sdk-support']['property-function'] || {}).js ? '>= ' + (prop['sdk-support']['property-function'] || {}).js : 'Not supported' %></td>
+      <td><%- (prop['sdk-support']['property-function'] || {}).ios ? '>= ' + (prop['sdk-support']['property-function'] || {}).ios : 'Not supported' %></td>
+      <td><%- (prop['sdk-support']['property-function'] || {}).android ? '>= ' + (prop['sdk-support']['property-function'] || {}).android : 'Not supported' %></td>
+    </tr>
```
lucaswoj: Did you consider allowing arbitrary fields rather than hardcoding `basic` and `property-function`? This would allow us to specify more nuanced information about specific properties as needed. 

For example, `icon-image` might want to specify 

```
{
  'sdk-support': {
    'Basic support': ...
    'Data-driven styling support`: ...
    'SDF support:' ...
  }
}
```

jfirebaugh: I definitely want to keep it extensible if we decide there are other relevant rows. However I think we want to keep the key in the JSON decoupled from the terminology in the docs. I can see the key in the JSON becoming relied on by other software (e.g. Studio), so we'd want to keep it stable, while at the same time we might want to change the terminology presented to the end user. So here for example I used `property-function` to match other parts of the JSON, but branded it "Data-driven styling" in the docs. And if this mapping expands beyond two rows, we'd certainly want to make the template more generic.

lucaswoj: 👍

lucaswoj:
:shipit:

PR #467, opened 2016-07-10
fix type links in docs

andrewharvey:
The type links in the docs eg. the link at https://www.mapbox.com/mapbox-gl-style-spec/#layer-filter is currently pointing to #filter but the anchor is actually #types-filter. This fixes this.


jfirebaugh:
Thanks!

PR #468, opened 2016-07-13
Document the supported range of geometric coordinates in vector tiles

lucaswoj:
cc @jfirebaugh @flippmoke @springmeyer 

ref https://github.com/mapbox/vector-tile-spec/issues/82
ref https://github.com/mapbox/mapbox-gl-js/pull/2858

### DIFF
--- a/docs/_generate/index.html
+++ b/docs/_generate/index.html
@@ -181,7 +181,7 @@ navigation:
           <h3 class='space-bottom1'><a href='#sources-vector' title='link to vector'>vector</a></h3>
           <p>
             A vector tile source. Tiles must be in <a href="https://www.mapbox.com/developers/vector-tiles/">Mapbox
-            Vector Tile format</a>. All geometric coordinates in vector tiles must be in the range <code>[-extent, extent * 2)</code> All layers that use a vector source must specify a <code>"source-layer"</code> value.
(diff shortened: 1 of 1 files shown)


docs/_generate/index.html
```diff
@@ -181,7 +181,7 @@
           <h3 class='space-bottom1'><a href='#sources-vector' title='link to vector'>vector</a></h3>
           <p>
             A vector tile source. Tiles must be in <a href="https://www.mapbox.com/developers/vector-tiles/">Mapbox
-            Vector Tile format</a>. All layers that use a vector source must specify a <code>"source-layer"</code> value.
+            Vector Tile format</a>. All geometric coordinates in vector tiles must be in the range <code>[-extent, (extent * 2) - 1)</code> All layers that use a vector source must specify a <code>"source-layer"</code> value.
```
jfirebaugh: Isn't the range here inclusive on both ends, due to the `- 1`?

lucaswoj: Ah! Yes! Good catch.

jingsam:
Is it means that I should regenerate tiles by reducing buffer size to solve the problem with [line rendering problem](https://github.com/mapbox/mapbox-gl-js/issues/2854)?

PR #471, opened 2016-07-19
Update GeoJSON source default values

mourner:
The default values in the spec were outdated since https://github.com/mapbox/mapbox-gl-js/commit/e745190838edf673151bc5f75869a0b0bb9c6fe8

cc @lucaswoj

### DIFF
--- a/reference/v8.json
+++ b/reference/v8.json
@@ -157,18 +157,18 @@
     },
     "maxzoom": {
       "type": "number",
-      "default": 14,
+      "default": 18,
       "doc": "Maximum zoom level at which to create vector tiles (higher means greater detail at high zoom levels)."
     },
     "buffer": {
       "type": "number",
-      "default": 64,
-      "doc": "Tile buffer size on each side (higher means fewer rendering artifacts near tile edges but slower performance)."
+      "default": 128,
+      "doc": "Tile buffer size on each side (relative to a 512px tile; higher means fewer rendering artifacts near tile edges but slower performance)."
     },
     "tolerance": {
       "type": "number",
-      "default": 3,
-      "doc": "Douglas-Peucker simplification tolerance (higher means simpler geometries and faster performance)."
+      "default": 0.375,
+      "doc": "Simplification tolerance (higher means simpler geometries and faster performance)."
     },
     "cluster": {
       "type": "boolean",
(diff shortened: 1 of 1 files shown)


reference/v8.json
```diff
@@ -157,18 +157,18 @@
...
-      "doc": "Tile buffer size on each side (higher means fewer rendering artifacts near tile edges but slower performance)."
+      "default": 128,
+      "doc": "Tile buffer size on each side (relative to a 512px tile; higher means fewer rendering artifacts near tile edges but slower performance)."
     },
     "tolerance": {
       "type": "number",
-      "default": 3,
-      "doc": "Douglas-Peucker simplification tolerance (higher means simpler geometries and faster performance)."
+      "default": 0.375,
+      "doc": "Simplification tolerance (higher means simpler geometries and faster performance)."
```
lucaswoj: Why did you remove "Douglas-Peucker" here?

reference/v8.json
```diff
@@ -157,18 +157,18 @@
...
-      "default": 14,
+      "default": 18,
       "doc": "Maximum zoom level at which to create vector tiles (higher means greater detail at high zoom levels)."
     },
     "buffer": {
       "type": "number",
-      "default": 64,
-      "doc": "Tile buffer size on each side (higher means fewer rendering artifacts near tile edges but slower performance)."
+      "default": 128,
+      "doc": "Tile buffer size on each side (relative to a 512px tile; higher means fewer rendering artifacts near tile edges but slower performance)."
```
lucaswoj: Does "relative to a 512px tile" mean that the value of `buffer` is specified in pixels, as if associated with a 512px tile?

reference/v8.json
```diff
@@ -177,8 +177,8 @@
     },
     "clusterRadius": {
       "type": "number",
-      "default": 400,
-      "doc": "Radius of each cluster when clustering points, relative to 4096 tile."
+      "default": 50,
+      "doc": "Radius of each cluster when clustering points, relative to 512px tile."
```
lucaswoj: Should we say that this radius is measured in pixels?

reference/v8.json
mourner: I think it's an implementation detail that's not important to most users, but sounds scary.

reference/v8.json
mourner: Yes. I just thought that simply mentioning that it's measured in pixels is not enough since it's not immediately apparent what a pixel is for a vector tile.

PR #477, opened 2016-08-05
Line break even if there's not a space

xrwang:
Add an attribute to control the way line breaks happen for CJK languages. `text-break-all` breaks according to a different set of rules. Refs https://github.com/mapbox/mapbox-gl-js/tree/cjk-linebreak

cc @nickidlugash @amyleew

### DIFF
--- a/reference/v8.json
+++ b/reference/v8.json
@@ -899,13 +899,13 @@
         }
       }
     },
-    "text-break-space": {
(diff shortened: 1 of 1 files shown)


xrwang:
Closing as this in lieu of using the unicode line breaking algorithm.

PR #482, opened 2016-08-24
Specify bearing and pitch direction

1ec5:
Amended the bearing and pitch property documentation to describe some reference points, based on how Mapbox GL interprets these properties.

/cc @lucaswoj


lucaswoj:
Looks great! Thanks for improving the docs. 🚢

PR #490, opened 2016-09-08
Ignore layout and paint properties for invalid layers

scothis:
Currently, if a style contains an invalid layer type and either a paint
or layout property for that layer, the style validator will throw.
Instead of throwing, these properties should be ignored, the layer type
will still be reported as an error.


jfirebaugh:
👍

PR #492, opened 2016-09-13
Clarify cluster options documentation

lucaswoj:
ref https://github.com/mapbox/mapbox-gl-native/issues/6236#issuecomment-246768910

### DIFF
--- a/reference/v8.json
+++ b/reference/v8.json
@@ -163,7 +163,9 @@
     "buffer": {
       "type": "number",
       "default": 128,
-      "doc": "Tile buffer size on each side (measured in 1/512ths of a tile; higher means fewer rendering artifacts near tile edges but slower performance)."
+      "maximum": 512,
+      "minimum": 0,
+      "doc": "Size of the tile buffer on each side. A value of 0 produces no buffer. A value of 512 produces a buffer as wide as the tile itself. Larger values produce fewer rendering artifacts near tile edges and slower performance."
     },
     "tolerance": {
       "type": "number",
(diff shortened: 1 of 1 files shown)


reference/v8.json
```diff
@@ -163,7 +163,9 @@
     "buffer": {
       "type": "number",
       "default": 128,
-      "doc": "Tile buffer size on each side (measured in 1/512ths of a tile; higher means fewer rendering artifacts near tile edges but slower performance)."
+      "maximum": 512,
```
1ec5, line 166: Isn’t the size of a tile configurable?

lucaswoj: All tiles are normalized to the same size when they are loaded into GL: 512 "units" wide and tall.

The size of a buffer is limited by our representation of a coordinate as a `short`. In practice, this means that buffers cannot be larger than 512 "units." If a tile has a buffer larger than 512 "units" we drop all coordinates outside of the supported range and log warning messages.

1ec5: > All tiles are normalized to the same size when they are loaded into GL: 512 "units" wide and tall.

This normalization should be documented somewhere, either as a caveat to `tileSize` or in the documentation string for this property, since the relationship between the two properties is non-obvious.

lucaswoj: There are no sources that accept both `tileSize` and `buffer` options. 

> Size of the tile buffer on each side. A value of 0 produces no buffer. A value of 512 produces a buffer as wide as the tile itself. Larger values produce fewer rendering artifacts near tile edges and slower performance.

I worded this documentation to avoid the need to explain tile resizing. The units are, for practical purposes, opaque and are explained as such.

PR #493, opened 2016-09-13
Add an explicit enum value to the spec for text and icon pitch and rotation

boundsj:
This implements the solution style spec updates described in https://github.com/mapbox/mapbox-gl-native/pull/6253#issuecomment-244856555

> one possible approach to address the special-caseness of these properties would be to add an explicit enum value to the spec like auto or inherit for text-rotation-alignment, icon-rotation-alignment and text-pitch-alignment.
> - Add auto enum value
> - Set default value to auto

cc @lbud @yhahn @jfirebaugh @1ec5


jfirebaugh:
Rather than change the value of the "basic" level of "sdk-support", I think we should add a separate key "'auto' value". And support [here](https://github.com/mapbox/mapbox-gl-style-spec/blob/885e2bdcaa715915bb7b97132008db9a6c90e950/docs/_generate/item.html#L85-L86) for rendering arbitrary keys from "sdk-support".

reference/v8.json
```diff
@@ -790,15 +791,19 @@
       "property-function": true,
       "values": [
         "map",
-        "viewport"
+        "viewport",
+        "auto"
       ],
-      "doc": "Aligns text to the plane of the `viewport` or the `map` when the map is pitched. Matches `text-rotation-alignment` if unspecified.",
+      "default": "auto",
+      "doc": "Aligns text to the plane of the `viewport` or the `map` when the map is pitched. Matches `text-rotation-alignment` if auto.",
```
1ec5: Nit: surround `auto` in backticks. (That way it’ll become `MGLTextPitchAlignmentAuto` when the iOS/macOS SDK documentation is automatically generated from this file.)

boundsj:
>  I think we should add a separate key "'auto' value". And support here for rendering arbitrary keys from "sdk-support".

This is done in https://github.com/mapbox/mapbox-gl-style-spec/pull/493/commits/4b8bfbd0301a822de57934085433dfbb401b1229. I used the expected native platform versions that `auto` will be supported in. I stubbed `js` as `TBD` since I'm not certain how the this change and the next js release that supports it will be orchestrated.

Also, the `auto` string where present in the `doc` value gained backticks for the benefit of ios/macos documentation

cc @jfirebaugh @lucaswoj @lbud @1ec5

reference/v8.json
```diff
@@ -790,15 +796,24 @@
       "property-function": true,
       "values": [
         "map",
-        "viewport"
+        "viewport",
+        "auto"
       ],
-      "doc": "Aligns text to the plane of the `viewport` or the `map` when the map is pitched. Matches `text-rotation-alignment` if unspecified.",
+      "default": "auto",
+      "doc": "Aligns text to the plane of the `viewport` or the `map` when the map is pitched. Matches `text-rotation-alignment` if `auto`.",
```
1ec5: While you're here, I think the use of backticks here would contribute to confusing documentation in the iOS and macOS SDKs. It'll say something like "Aligns text to the plane of the `MGLTextPitchAlignmentViewport` or the `MGLTextPitchAlignmentMap` when the map is pitched." But the intent is to say "plane of the viewport when set to `viewport`".

1ec5 (changes requested):

reference/v8.json
boundsj: Yeah I see what you mean. In 5ab221e I updated so it now reads:

#### spec docs

screen shot 2016-09-15 at 9 14 40 pm

#### iOS SDK docs

screen shot 2016-09-15 at 9 14 27 pm

1ec5 (approved):

reference/v8.json
```diff
@@ -607,6 +608,11 @@
           "js": "0.10.0",
           "ios": "2.0.0",
           "android": "2.0.1"
+        },
+        "auto": {
+          "js": "TBD",
```
lucaswoj: I assume "TBD" stands for "to be decided." Why "TBD" instead of leaving this `undefined` to indicate that the feature is not supported?

reference/v8.json
```diff
@@ -595,10 +595,11 @@
       "property-function": true,
       "values": [
         "map",
-        "viewport"
+        "viewport",
+        "auto"
       ],
-      "default": "viewport",
-      "doc": "Orientation of icon when map is rotated.",
+      "default": "auto",
```
lucaswoj, line 601: Where do we explain what `auto` means?

lucaswoj (changes requested):

reference/v8.json
boundsj: Yeah that was my question in https://github.com/mapbox/mapbox-gl-style-spec/pull/493#issuecomment-247483816. Sounds like the process is to land this PR first and leave all 3 SDKs as `undefined`. Then, land the SDK PRs to support `auto`, then come back here and make a new PR to update support. Is my understanding correct?

reference/v8.json
boundsj: @lucaswoj do the changes to the doc strings in 296d335 address your concerns about the lack of explanation of `auto`?

reference/v8.json
boundsj: @lucaswoj 296d335 makes support for all SDKs for undefined for `auto` for now. I'm guessing about how a style spec change this should all be orchestrated to the SDKs so please let me know when we should add the specific version (at least on the GL JS side). 

https://github.com/mapbox/mapbox-gl-js/pull/3203 has added support for `auto` to GL JS but still needs review and the tests in https://github.com/mapbox/mapbox-gl-test-suite/pull/144

lucaswoj: > let me know when we should add the specific version (at least on the GL JS side)

Our policy is to update the style spec once the feature goes out in a public release. The release coordinator is responsible for this update.

reference/v8.json
lucaswoj: Looks good! Thanks 😄

lucaswoj (approved):

jfirebaugh (changes requested):
I want to suggest a slightly different way of doing the SDK support matrix -- will push to a separate branch for discussion.

jfirebaugh:
https://github.com/mapbox/mapbox-gl-style-spec/compare/symbol-complex-default...symbol-complex-default-alt
- Supports markdown in 1st column
- Custom rows appear after "Base support" and "Data-driven styling"

image

If you are 👍  please pull into this branch and merge.

lucaswoj:
Why "Base _support_" but not "Data-driven styling _support_" or "`auto` value _support_"?

jfirebaugh:
@lucaswoj This PR doesn't change that; let's not nitpick it here.

boundsj:
@jfirebaugh 👍 I merged the commits from your branch

jfirebaugh (approved):

Issue #456, opened 2016-05-26
3d extrusion (buildings etc) spec


lbud:
Spinning off of https://github.com/mapbox/mapbox-gl-js/issues/684#issuecomment-221713082: I've reimplemented @ansis's 3d building demo ([js branch](https://github.com/mapbox/mapbox-gl-js/tree/building-redux); [spec branch](https://github.com/mapbox/mapbox-gl-style-spec/tree/buildings); [shader branch](https://github.com/mapbox/mapbox-gl-shaders/tree/js-buildings)). That code is ported pretty faithfully from the demo, except that I've renamed `building` to `extrusion` everywhere: it 
- adds a new `extrusion` type (in which one would specify a polygon source layer)
- hardcodes a light direction
- extrudes a height based on a feature's `properties.levels` (the demo is done based on a source from @ajashton with this property), or falls back to a hardcoded `levels` integer
- I also added a `extrusion-shadow-color`, which colors the shadow tint (but is pretty subtle; doesn't change the amount of possible shadowing)

I propose the following additions to the spec (:warning: not all of which are yet implemented):
- We add a new `extrusion` type. At present it _only_ extrudes polygons. We may add support for extruding lines and points in the future, given compelling use cases/need.
- Extrusion properties:
  - layout:
    - `visibility`
    - Presumably there will be additional layout properties concerning label placement etc once more cases are tested
  - paint:
    - `extrusion-color`
    - `extrusion-shadow-color` (open to other naming suggestions -- @tmcw might have one)
    - `extrusion-opacity` (I'm on the fence about this as it can create [weird ghostly buildings](https://twitter.com/lbudorick/status/735603614916444162), but can support if people find ghostly buildings compelling)
    - `extrusion-pattern`: implementation has been punted so far but I fear this should probably exist
    - `extrusion-antialias`
    - `extrusion-outline-color` (and if so, what should an extrusion outline be? Just the base? Other edges? (Hint: I haven't done other edges))
    - `extrusion-translate`
    - `extrusion-translate-anchor`
    - `extrusion-height`: Though undoubtably there are cases to extrude not based on data (@1ec5 would here point out that the buildings in DC are more or less all the same height _by law_), **I think it's important this new type be DDS-first**.
    - `extrusion-lighting-anchor` (whether a lighting direction should be oriented to map or viewport)
- APIs:
  - `map.setLightDirection` to change lighting

@mapbox/gl @mapbox/cartography-cats

[...]

PR #495, opened 2016-09-16
Add fill-extrude-height, fill-extrude-base, and light root property

lbud:
This
- adds `fill-extrude-height` and `fill-extrude-base` as fill layer paint properties
- adds a `light` root object with the following properties, for use in lighting extruded features:
  - `anchor`
  - `color`
  - `direction`
  - `intensity`

Part of https://github.com/mapbox/mapbox-gl-js/issues/684, result of discussion on https://github.com/mapbox/mapbox-gl-style-spec/issues/456.

cc @lucaswoj @jfirebaugh @mourner @mollymerp @scothis @nickidlugash


lib/diff.js
```diff
@@ -87,7 +87,12 @@ var operations = {
     /*
      * { command: 'setTransition', args: [transition] }
      */
-    setTransition: 'setTransition'
+    setTransition: 'setTransition',
+
+    /*
+     * { command: 'setLighting', args: [lightProperties] }
+     */
+    setLighting: 'setLighting'
```
lucaswoj: It looks like we use the name `light` rather than `lighting` elsewhere to refer to light properties. What do you think about changing this name to `setLight`?

reference/v8.json
```diff
@@ -43,6 +43,37 @@
...
+        ],
+        "doc": "Whether extruded geometries are lit relative to map or viewport. The light anchor property will only be used if it has not been set by map options upon initialization.",
+        "example": "map"
+      },
+      "direction": {
+        "type": "array",
+        "value": "number",
+        "default": [1.15, 210, 30],
+        "doc": "Light direction for lighting extruded geometries, in [r radial coordinate, a azimuthal angle, p polar angle] where r indicates the distance from the center of an object to its light, a indicates the position of the light relative to 0º (like a clock), and p indicates the height of the light (from 0º, directly above, to 180º, directly below). The light direction property will only be used if it has not been set by map options upon initialization.",
+        "example": [1.5, 90, 80]
```
lucaswoj, line 72: Do you think our code would be more self-documenting if this were split into 3 separate properties?

lib/diff.js
lbud: I'm fine with that — `lighting` was just instinctual but you're right that `light` would make more sense with the spec additions 👍

scothis: The name and arguments of the command should match the API in gl-js.

reference/v8.json
lbud: Maybe more self-documenting, but I hesitate because:
- the naming of spherical coordinates is [so fraught as it is](http://mathworld.wolfram.com/SphericalCoordinates.html), and I'd rather loosely suggest names in the spec docstring than encode a certain naming convention in every style, and
- (related) using an array kind of allows a user to forget their "names" — even in my mind already, arr[0] is the length, arr[1] is the clock angle, arr[2] is the light height. Having to use "radial," "azimuthal," and "polar" in the style or, worse, in a `setLight()` call, seems worse than supplying an array to "direction"

lib/diff.js
lbud: @scothis I [changed it](https://github.com/mapbox/mapbox-gl-js/blob/building_redux_renderbuffer_dds/js/ui/light.js#L60-L72) while you weren't looking 🙈

lib/diff.js
```diff
@@ -245,6 +250,27 @@ function diffLayers(before, after, commands) {
...
 
+function diffLight(before, after, commands) {
+    before = before || {};
+    after = after || {};
+
+    var lightProps = ['anchor', 'direction', 'color', 'intensity'];
+
+    var newProps = {};
+
+    for (var i in lightProps) {
```
scothis: Creating a minimal diff feels like overkill. We can check if `!isEqual(before, after)` and push a command for the entire `after` lighting object if so.

Iterating over each property was only necessary when the `setLighting` command accepted a single property per invocation.

reference/v8.json
scothis: Using an `[x, y, z]` array to describe a vector is very common. I'm less familiar with polar coordinates.

reference/v8.json
```diff
@@ -43,6 +43,37 @@
...
+          "viewport"
+        ],
+        "doc": "Whether extruded geometries are lit relative to map or viewport. The light anchor property will only be used if it has not been set by map options upon initialization.",
+        "example": "map"
+      },
+      "direction": {
+        "type": "array",
+        "value": "number",
+        "default": [1.15, 210, 30],
+        "doc": "Light direction for lighting extruded geometries, in [r radial coordinate, a azimuthal angle, p polar angle] where r indicates the distance from the center of an object to its light, a indicates the position of the light relative to 0º (like a clock), and p indicates the height of the light (from 0º, directly above, to 180º, directly below). The light direction property will only be used if it has not been set by map options upon initialization.",
```
1ec5: `º` is the masculine ordinal used for instance in the Spanish abbreviation “1º” (primero). The degree symbol is `°`. (On a Mac using the U.S. or Extended ABC keyboard layout, use ⇧⌥8 to access the degree symbol.)

reference/v8.json
```diff
@@ -43,6 +43,37 @@
...
+        "default": "viewport",
+        "values": [
+          "map",
+          "viewport"
+        ],
+        "doc": "Whether extruded geometries are lit relative to map or viewport. The light anchor property will only be used if it has not been set by map options upon initialization.",
+        "example": "map"
+      },
+      "direction": {
+        "type": "array",
```
1ec5, line 63: Specify `"length": 3`.

reference/v8.json
1ec5: I think it would be more intuitive to describe this property as the position of the light source relative to the feature, rather than the lighting direction. “Lighting direction” could be confused with the direction to which the light source points, as opposed to the position from which it comes. (Or am I misunderstanding this property?)

reference/v8.json
1ec5: It is common to express polar coordinates as a tuple as well. However, if the concern is that most developers would be unfamiliar with this convention, we could use the terms `yaw`, `pitch`, and `roll` (or `heading`, `elevation`, and `bank`), which are better understood than “radial”, “azimuthal”, and “polar”.

Whatever the case, I think we should be very explicit in our documentation for these properties, similar to [Apple’s documentation for a similar construct](https://developer.apple.com/reference/avfoundation/avaudio3dangularorientation?language=objc).

reference/v8.json
```diff
@@ -43,6 +43,37 @@
...
+        "type": "color",
+        "default": "#ffffff",
+        "doc": "Color tint for lighting extruded geometries. The light color property will only be used if it has not been set by map options upon initialization."
+      },
+      "intensity": {
+        "type": "number",
+        "maximum": 1,
+        "minimum": 0,
+        "default": 0.5,
+        "doc": "Intensity of lighting (on a scale from 0 to 1). Higher numbers will present as more extreme contrast. The light intensity property will only be used if it has not been set by map options upon initialization."
```
1ec5: We should spell out the visual effect of the default value of 0.5. (By the way, [this Apple documentation](https://developer.apple.com/reference/scenekit/scnlight/1640548-intensity) may be helpful in figuring out the right way to discuss light intensity for physical versus non-physical lighting.)

reference/v8.json
```diff
@@ -1439,6 +1470,31 @@
...
       }
     },
+    "fill-extrude-height": {
+      "type": "number",
+      "function": "interpolated",
+      "zoom-function": true,
+      "property-function": true,
+      "default": 0,
+      "minimum": 0,
+      "doc": "The height with which to extrude this fill layer.",
```
1ec5, line 1698: Specify `"units": "meters"`.

reference/v8.json
```diff
@@ -1439,6 +1470,31 @@
...
+      "transition": true
+    },
+    "fill-extrude-base": {
+      "type": "number",
+      "function": "interpolated",
+      "zoom-function": true,
+      "property-function": true,
+      "default": 0,
+      "minimum": 0,
+      "doc": "The height with which to extrude the base of this layer.",
```
1ec5, line 1708: Specify `"units": "meters"`.

reference/v8.json
```diff
@@ -43,6 +43,37 @@
...
     },
+    "light": {
+      "anchor": {
+        "type": "enum",
+        "default": "viewport",
+        "values": [
+          "map",
+          "viewport"
+        ],
+        "doc": "Whether extruded geometries are lit relative to map or viewport. The light anchor property will only be used if it has not been set by map options upon initialization.",
```
1ec5: Should this be `light-anchor` or `light.anchor`? Should be in backticks in any case.

1ec5: “relative to **the** map”

1ec5 (changes requested):
Looking good. Just a few nits related to documentation.

reference/v8.json
```diff
@@ -43,6 +43,38 @@
...
+        ],
+        "doc": "Whether extruded geometries are lit relative to the map or viewport. The `light.anchor` property will only be used if it has not been set by map options upon initialization.",
+        "example": "map"
+      },
+      "direction": {
+        "type": "array",
+        "value": "number",
+        "length": 3,
+        "default": [1.15, 210, 30],
+        "doc": "Position of the light source relative to lit (extruded) geometries, in [r radial coordinate, a azimuthal angle, p polar angle] where r indicates the distance from the center of an object to its light, a indicates the position of the light relative to 0° (like a clock), and p indicates the height of the light (from 0°, directly above, to 180°, directly below). The light direction property will only be used if it has not been set by map options upon initialization.",
```
1ec5: - [ ] Is the distance in meters?
- [ ] Which direction does a = 0° point? (Also, clockwise, not "like a clock".)
- [ ] Is the distance and angle measured from the volumetric center of the extruded geometry, or from the surface?

1ec5 (changes requested):

reference/v8.json
```diff
@@ -43,6 +43,38 @@
...
+        "type": "enum",
+        "default": "viewport",
+        "values": [
+          "map",
+          "viewport"
+        ],
+        "doc": "Whether extruded geometries are lit relative to the map or viewport. The `light.anchor` property will only be used if it has not been set by map options upon initialization.",
+        "example": "map"
+      },
+      "direction": {
```
1ec5: Given the current documentation, I think you should strongly consider renaming the property to "position", even if the _goal_ of setting this property is to change the lighting direction. (You can say as much in the documentation.)

1ec5 (changes requested):

reference/v8.json
```diff
@@ -43,6 +43,42 @@
...
       "example": 50
     },
+    "light": {
+      "anchor": {
+        "type": "enum",
+        "default": "viewport",
+        "values": [
+          "map",
+          "viewport"
+        ],
```
davidtheclark: The format for enum values changed in #510. 😬

reference/v8.json
```diff
@@ -1439,6 +1475,33 @@
...
+      "zoom-function": true,
+      "property-function": true,
+      "default": 0,
+      "minimum": 0,
+      "units": "meters",
+      "doc": "The height with which to extrude the base of this layer.",
+      "transition": true,
+      "requires": [
+        {
+          "<=": "fill-extrude-height"
```
davidtheclark, line 1712: This `<=` key to a `requires` object is new, right? So its introduction will require changes to documentation generation code and Studio.

reference/v8.json
lbud: The radial coordinate doesn't have an equivalent in `yaw, pitch, roll`. Moreover, those descriptions of a person's head don't match up in my mind to descriptions of a light source _outside_ someone's head.

I'm open to improvements to this spec description, but I think we should merge this for now.

reference/v8.json
lbud: done

lib/diff.js
lbud: changed

lib/diff.js
lbud: /giphy grool
done

lbud:
ok ready ready ready
@1ec5 👍 ?

reference/v8.json
```diff
@@ -68,7 +68,7 @@
         "function": "interpolated",
         "zoom-function": true,
         "property-function": false,
-        "doc": "Position of the light source relative to lit (extruded) geometries, in [r radial coordinate, a azimuthal angle, p polar angle] where r indicates the distance from the center of an object to its light, a indicates the position of the light relative to 0° (like a clock), and p indicates the height of the light (from 0°, directly above, to 180°, directly below).",
+        "doc": "Position of the light source relative to lit (extruded) geometries, in [r radial coordinate, a azimuthal angle, p polar angle] where r indicates the distance from the center of an object to its light, a indicates the position of the light relative to 0° (0° corresponds to 12:00 on a clock, and degrees proceed clockwise; 0° with `light.anchor: viewport` corresponds to the top of the viewport, or with `light.anchor: map` to due north), and p indicates the height of the light (from 0°, directly above, to 180°, directly below).",
```
1ec5: “12:00 on a clock” is no longer necessary, because “top of the viewport” and “due north” are descriptive enough.

Replace:

> with `light.anchor: viewport`

with:

> when `light.anchor` is set to `viewport`

The iOS and macOS SDKs have a script that translates the construct to something more appropriate for the language, so it’s best to use it when possible.

1ec5 (changes requested):

1ec5 (approved):
🎈

lucaswoj (approved):
Let's do this!

PR #501, opened 2016-09-20
Make SDK support terminology more consistent

### LINKED ISSUE #498: Create consistent SDK support naming convention
Why "Base _support_" but not "Data-driven styling _support_" or "`auto` value _support_"?

Some good examples to follow:
- http://node.green/
- http://caniuse.com/#search=es6

lucaswoj:
This PR makes SDK support terminology more consistent. All terms complete the sentence "The SDK supports ____"). 

This PR also eliminates presentational renaming of SDK support terminology (i.e. `property-function` -> `data-driven styling`). This makes the codebase less surprising. Although the strings are human-readable we should treat their modification as a breaking API change because they may be relied upon by external software.

resolves #498 
related to https://github.com/mapbox/mapbox-gl-style-spec/pull/465#discussion-diff-69342567

screen shot 2016-09-20 at 4 47 33 pm
screen shot 2016-09-20 at 4 49 31 pm

cc @jfirebaugh @boundsj @lbud @mollymerp @mapsam

### DIFF
--- a/docs/_generate/index.html
+++ b/docs/_generate/index.html
@@ -231,7 +231,7 @@ navigation:
             </thead>
             <tbody>
             <tr>
-              <td>Base support</td>
+              <td>basic functionality</td>
               <td class='center'>&gt;= 0.10.0</td>
               <td class='center'>&gt;= 2.0.0</td>
               <td class='center'>&gt;= 2.0.1</td>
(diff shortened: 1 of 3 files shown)


jfirebaugh (approved):
Yeah, in retrospect it's not worth avoiding some use of display strings as keys if we don't avoid them completely. This keeps things simpler.

PR #503, opened 2016-09-22
Clarify descriptions of icon-text-fit and icon-text-fit-padding

davidtheclark:
I was reading these descriptions and thought of some adjustments that clarified them for my mind, and might do so for others.

### DIFF
--- a/reference/v8.json
+++ b/reference/v8.json
@@ -646,7 +646,7 @@
         "height"
       ],
       "default": "none",
-      "doc": "Position and scale of an icon relative to the dimensions of text in `text-field`.",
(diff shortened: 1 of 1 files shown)


reference/v8.json
```diff
@@ -646,7 +646,7 @@
         "height"
       ],
       "default": "none",
-      "doc": "Position and scale an icon by the its corresponding text.",
+      "doc": "Position and scale of an icon relative to the dimensions of text in `text-field`.",
```
jfirebaugh: I find that enumerated properties are best documented by providing one sentence about what the property does in general, and then descriptions of the effects of each value. For example:

> Permits scaling the icon such that it fits the associated text. The value `none` performs no scaling. The values `width` and `height` produce an icon scaled in the x- or y-dimension, respectively, such that the text fits inside the icon. The value `both` produces scaling in both dimensions.

davidtheclark: Good suggestion. Will amend.

jfirebaugh (approved):

PR #505, opened 2016-09-23
Clarify icon-text-fit-padding's icon-text-fit requirement

### LINKED ISSUE #504: Allow one property to require another property to have an array of possible values
There's a problem with `icon-text-fit-padding`. It [says that it requires `icon-text-fit`](https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/reference/v8.json#L677), but in fact it requires `icon-text-fit` to have any of its possible values other than `none`. With `"icon-text-fit": "none"`, `icon-text-fit-padding` is no good.

Looks like [there is existing syntax](https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/reference/v8.json#L452) for one property requiring that another property have _a specific value_; but not for this case, where one property requires that another property have _one of several values_.

Can we add the ability for an object within `"requires"` to take an array of possibilities as its value?

For example, to address the root issue with `icon-text-fit-padding`:

``` json
"requires": [ 
  "icon-image",
  {
    "icon-text-fit": ["both", "width", "height"]
  },
  "text-field"
]
```

cc @samanpwbb

davidtheclark:
Closes #504.
- Specifies the particular `icon-text-fit` values that are required by `icon-text-fit-padding`, introducing a new possibility for the `requires` property:  arrays as required property values.
- Modifies doc generation code to accommodate the above.

This template code renders like so:

screen shot 2016-09-23 at 12 05 59 pm

### DIFF
--- a/docs/_generate/item.html
+++ b/docs/_generate/item.html
@@ -32,7 +32,11 @@
             if (req['!']) {
               return '<em>Disabled by</em> <var>' + req['!'] + '</var>.';
             } else {
-              return '<em>Requires</em> <var>' + _.toPairs(req)[0][0] + ' = ' + _.toPairs(req)[0][1] + '</var>.';
+              var requiredValue = _.toPairs(req)[0][1];
+              var requiredDisplay = _.isArray(requiredValue)
+                ? '</var>one of<var> ' + requiredValue.join('</var>,<var> ')
+                : requiredValue;
+              return '<em>Requires</em> <var>' + _.toPairs(req)[0][0] + ' = ' + requiredDisplay + '</var>.';
             }
           }
         }).join(' ') %>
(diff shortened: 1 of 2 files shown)


docs/_generate/item.html
```diff
@@ -32,7 +32,11 @@
             if (req['!']) {
               return '<em>Disabled by</em> <var>' + req['!'] + '</var>.';
             } else {
-              return '<em>Requires</em> <var>' + _.toPairs(req)[0][0] + ' = ' + _.toPairs(req)[0][1] + '</var>.';
+              var requiredValue = _.toPairs(req)[0][1];
+              var requiredDisplay = (_.isArray(requiredValue))
```
jfirebaugh: Nit: no need for outer parentheses.

jfirebaugh (approved):

PR #510, opened 2016-09-28
fixes #497: document enum values

### LINKED ISSUE #497: Document enum values individually
Change:

```
    "icon-rotation-alignment": {
      "values": [
        "map",
        "viewport",
        "auto"
      ],
      "default": "auto",
      "doc": "Orientation of icon when map is rotated. Aligns icon to the plane of the viewport when set to `viewport` or the plane of the map when set to `map`. Selecting `auto` defaults to `map` for line placement and `viewport` for symbol placement",
```

To:

```
    "icon-rotation-alignment": {
      "values": {
        "map": {
          "doc": "Icons are oriented north-up, regardless of map rotation."
        },
        "viewport": {
          "doc": "Icons are oriented such that they are always upright on screen."
        },
        "auto": {
          "doc": "Equivalent to `map` if this layer uses `"symbol-placement": "line"`, otherwise `viewport`."
        }
      },
      "default": "auto",
      "doc": "Determines the orientation behavior of icons when the map is rotated.",
```

Then the generated docs can list the enum values with their documentation in a definition list.

incanus:
Affects a bunch of downstream stuff which is being documented in #497 proper. 

Also removes color operations in https://github.com/mapbox/mapbox-gl-style-spec/pull/510/commits/44da3a8b5e06b1fa7aea4313ab71a9c358fb0d57 per chat with @samanpwbb since these were cut back in https://github.com/mapbox/mapbox-gl-style-spec/issues/308. 

/cc @jfirebaugh @1ec5


reference/v8.json
```diff
@@ -596,13 +650,19 @@
...
       "zoom-function": true,
       "property-function": true,
-      "values": [
-        "map",
-        "viewport",
-        "auto"
-      ],
+      "values": {
+        "map": {
+            "doc": "When `symbol-placement` is set to `point`, produces icons whose x-axes are aligned east-west."
```
jfirebaugh: What about symbol-placement = line?

reference/v8.json
```diff
@@ -823,13 +897,19 @@
...
       "zoom-function": true,
       "property-function": true,
-      "values": [
-        "map",
-        "viewport",
-        "auto"
-      ],
+      "values": {
+        "map": {
+            "doc": "When `symbol-placement` is set to `point`, produces glyphs whose x-axes are aligned east-west."
```
jfirebaugh: Ditto.

reference/v8.json
```diff
@@ -1241,43 +1359,64 @@
...
-      "!in",
-      "all",
-      "any",
-      "none",
-      "has",
-      "!has"
-    ],
+    "values": {
+      "==": {
+          "doc": "The equal-to operator."
```
jfirebaugh: Let's pull in the descriptions from https://www.mapbox.com/mapbox-gl-style-spec/#types-filter for these.

jfirebaugh (changes requested):

reference/v8.json
```diff
@@ -485,10 +535,14 @@
...
+      "values": {
+          "point": {
+              "doc": "The label is placed at the point where the geometry is located."
+          },
+          "line": {
+              "doc": "The label is placed along the line of the geometry."
+          }
+      },
       "default": "point",
       "doc": "Label placement relative to its geometry. `line` can only be used on LineStrings and Polygons.",
```
1ec5: Let’s move the second sentence to the `line` documentation.

reference/v8.json
```diff
@@ -799,13 +867,19 @@
...
-      ],
+      "values": {
+        "map": {
+            "doc": "The text is aligned to the plane of the map."
+        },
+        "viewport": {
+            "doc": "The text is aligned to the plane of the viewport."
+        },
+        "auto": {
+            "doc": "Set to the same value as `text-rotation-alignment`."
```
1ec5: > Automatically matches the value of `text-rotation-alignment`.

reference/v8.json
```diff
@@ -983,17 +1069,35 @@
...
+            "doc": "The right side of the text is placed closest to the anchor."
+        },
+        "top": {
+            "doc": "The top of the text is placed closest to the anchor."
+        },
+        "bottom": {
+            "doc": "The bottom of the text is placed closest to the anchor."
+        },
+        "top-left": {
+            "doc": "The top left side of the text is placed closest to the anchor."
```
1ec5: s/top left side/top-left corner/

reference/v8.json
```diff
@@ -1098,11 +1202,17 @@
...
-        "none",
-        "uppercase",
-        "lowercase"
-      ],
+      "values": {
+        "none": {
+            "doc": "The text is not altered."
+        },
+        "uppercase": {
+            "doc": "The text is uppercased."
```
1ec5: > Forces all letters to be displayed in uppercase.

This way there’s no ambiguity about `capitalize`-style behavior.

reference/v8.json
```diff
@@ -324,12 +346,16 @@
...
+        "visible": {
+            "doc": "The layer is shown."
+        },
+        "none": {
+            "doc": "The layer is not shown."
+        }
+      },
       "default": "visible",
-      "doc": "The display of this layer. `none` hides this layer.",
+      "doc": "The display of this layer.",
```
1ec5: > Whether this layer is displayed.

reference/v8.json
incanus: Whoops, de06217e5bce816a6dc895ab782c12529a166c5f.

reference/v8.json
incanus: de06217e5bce816a6dc895ab782c12529a166c5f

reference/v8.json
incanus: Do you mean literally, as in `set inclusion: feature[key] ∈ {v0, ..., vn}`?

reference/v8.json
```diff
@@ -1440,11 +1585,15 @@
...
       "zoom-function": true,
       "property-function": true,
-      "values": [
-        "map",
-        "viewport"
-      ],
-      "doc": "Control whether the translation is relative to the map (north) or viewport (screen)",
+      "values": {
+        "map": {
+            "doc": "The fill is translated relative to the map (north)."
```
1ec5: What does “north” mean here? Does it mean that a higher y value moves the fill’s origin to the north?

reference/v8.json
```diff
@@ -1440,11 +1585,15 @@
...
+        },
+        "viewport": {
+            "doc": "The fill is translated relative to the viewport (screen)."
```
1ec5: What does “screen” mean here? Does it mean that a higher y value moves the fill’s origin up? Or down? (I suppose specifying one or the other would be problematic for platform-specific documentation such as mapbox/mapbox-gl-native#6066, but that isn’t the style specification’s problem.)

1ec5 (changes requested):

reference/v8.json
incanus: How about db471343f90b4fe32b61c802ae99e5cfe72b1926?

1ec5: Another option is to use English, for example:

> A feature is included when the specified property is one of the specified values.

1ec5: I think the problem with just saying “equal-to operator” is that it isn’t clear that the left-hand side must be the property name and the right-hand side must be a property value. Even saying “left” and “right” could be confusing, though, given that the JSON representation is `["==", "key", "value"]` rather than `["key", "==", "value"]`.

samanpwbb:
In order to shorten the descriptions, can we cut the "when set to {val}" part of all the descriptions? I don't think it's necessary.

jfirebaugh:
They're all supposed to read "When `symbol-placement` is set to...", referring to the behavior of other properties in the presence of `symbol-placement`. There's a couple of places where "`symbol-placement` is" is omitted where that's unclear that we should fix, but we can't remove that text entirely.

reference/v8.json
```diff
@@ -823,13 +897,19 @@
...
       "zoom-function": true,
       "property-function": true,
-      "values": [
-        "map",
-        "viewport",
-        "auto"
-      ],
+      "values": {
+        "map": {
+            "doc": "When `symbol-placement` is set to `point`, produces glyphs whose x-axes are aligned east-west. When set to `line`, produces glyphs whose x-axes are aligned with the line."
```
samanpwbb: >  produces glyphs whose x-axes are aligned east-west

"produces glyphs who x-axes" is unnecessarily techincal. This could benefit from plainer language. 

Point: Aligns text east-west. 
Line: Places text along corresponding line geometry.

reference/v8.json
```diff
@@ -639,14 +699,22 @@
...
       "property-function": false,
-      "values": [
-        "none",
-        "both",
-        "width",
-        "height"
-      ],
+      "values": {
+        "none": {
+            "doc": "No scaling is performed."
```
jfirebaugh: "The icon is displayed at its intrinsic aspect ratio." (`icon-size` can still scale it.)

reference/v8.json
```diff
@@ -639,14 +699,22 @@
...
+        },
+        "both": {
```
jfirebaugh: Make this the last property, so they are documented in a natural order.

reference/v8.json
```diff
@@ -639,14 +699,22 @@
...
+            "doc": "Scaling is done in both x- and y-dimensions."
+        },
+        "width": {
+            "doc": "Scaling is done in the x-dimension to fit the text's dimensions."
```
jfirebaugh: "The icon is scaled in the x-dimension to fit the width of the text."

reference/v8.json
jfirebaugh: > Point: Aligns text east-west. 

👍 

> Line: Places text along corresponding line geometry.

There's an important nuance that this doesn't convey; it's the difference between:

```
    ╱
   ↗
  ╱
 ↗
╱
```

and

```
    ╱
   →
  ╱
 →
╱
```

reference/v8.json
incanus: 83aa844ba7d4089b11c1e6757389e84dd92c261b

reference/v8.json
incanus: 83aa844ba7d4089b11c1e6757389e84dd92c261b

reference/v8.json
incanus: 83aa844ba7d4089b11c1e6757389e84dd92c261b

reference/v8.json
incanus: 83aa844ba7d4089b11c1e6757389e84dd92c261b

reference/v8.json
incanus: 83aa844ba7d4089b11c1e6757389e84dd92c261b

reference/v8.json
incanus: 83aa844ba7d4089b11c1e6757389e84dd92c261b

reference/v8.json
incanus: 742bc7a72a9a00e7f5ec3369339a7f78cb2b85d2

incanus:
> There's a couple of places where "`symbol-placement` is" is omitted where that's unclear that we should fix

ca7b44676b7e95bce108fda1ec03bbd68df3a000

reference/v8.json
incanus: ab29a11fb0a02fec1a36c78c17391a33a80f1a7d

reference/v8.json
incanus: 6b8c8e250e4c6acbbe524d2fb97c852e2da8e6dd

reference/v8.json
incanus: 34771183e977f15951a1c7b26c0a9814f28b18d3

reference/v8.json
incanus: 6b79dae157978ed7d195edcf216c3fdd7137aa42

incanus:
~~Ready for re-review @1ec5 @jfirebaugh @samanpwbb.~~

Oops, still need to hit doc generator etc.

incanus:
Doc generator grabbed in 674a1e1e62ac37da7210e707afba870a10101a88. 

Do we want to update the page layout now after this? Now we have: 

screen shot 2016-09-28 at 4 11 44 pm

However, we don't yet capitalize on our new `enum` documentation. Should we in this context?

jfirebaugh:
Yes, let's stick them in a `<dl>`.

incanus:
Ok, ready for review again. Downstream work is happening in https://github.com/mapbox/mapbox-gl-native/pull/6508 and is looking good as well.

jfirebaugh (approved):

jfirebaugh:
image

incanus:
@1ec5 Mind re-reviewing and unblocking this?

reference/v8.json
```diff
@@ -1241,43 +1359,64 @@
...
-      "!in",
-      "all",
-      "any",
-      "none",
-      "has",
-      "!has"
-    ],
+    "values": {
+      "==": {
+          "doc": "The equality operator."
```
1ec5: Can we be any more specific? The equality operator ==, as most programmers know it, allows the left and right sides to be reversed. That isn't the case with any of these operators.

1ec5 (changes requested):

reference/v8.json
incanus: I think this is pretty clear in the context of GL styling, or if not, is very easily fixed (flip things, using the format `key == value`).

reference/v8.json
incanus: https://github.com/mapbox/mapbox-gl-style-spec/pull/510#discussion_r81393307

reference/v8.json
1ec5: I disagree. Equality is the most straightforward of the operators, but for example describing `has` as “the existence operator” doesn’t elucidate it for anyone who’s new to filters. How about something like this:

``` md
The inequality operator: <var>feature[key]</var> ≠ <var>value</var>.
```

jfirebaugh: Yes, that's what I meant by "pull in the descriptions from https://www.mapbox.com/mapbox-gl-style-spec/#types-filter". I think it would also be good to set up the `key` and `value` variables as the existing documentation does:

image

incanus: Trying this and testing downstream implications in https://github.com/mapbox/mapbox-gl-style-spec/pull/510/commits/5848d76dc10cc3ac16b395fff9e8df922eef2f3e.

incanus: Downstream isn't affected, since we handle filter operators manually in the SDKs. How do you feel about this @1ec5 @jfirebaugh?

1ec5 (approved):

PR #542, opened 2016-10-19
Fix light documentation

### LINKED ISSUE #522: “light” is undocumented
image

1ec5:
Moved the `light` documentation added in #495 to the right location so it shows up in the published HTML documentation.

property
☝️ _`light` root property_

type
☝️ _`Light` type_

Fixes #522.

/cc @lbud


1ec5:
Note that the link to the “light” type within the “light” root property is broken: #543.

lbud (approved):
🙌  thanks @1ec5! 🍏

PR #553, opened 2016-10-28
Add deref

jfirebaugh:
Preparation for https://github.com/mapbox/mapbox-gl-js/issues/1393.


mourner:
Are these utilities to be run from within GL JS on all style changes? What's the reason to start them in style-spec? To reuse in migration scripts?

jfirebaugh:
`deref`, yes, for migrations. It's looking like `group_by_layout` belongs better in gl-js though.

mourner (changes requested):
Looks great! Before we merge, let's remove `groupByLayout` since it moved to GL JS.

PR #558, opened 2016-11-04
Don't tolerate arbitrary keys anymore

lbud:
🔪 to arbitrary keys 🔪 
@lucaswoj — I then added a more specific message just for "interactive" in https://github.com/mapbox/mapbox-gl-style-spec/commit/6880747337c18cbf9c1c9c049adf88a34d24870f (`"layers[0]: unknown property \"interactive\"; the \"interactive\" key is deprecated"`) — 👍 / 👎 ?


lbud:
Per chat, going to go about this differently — we can't really predict how far-reaching the implications of this kind of breaking change will be. cc @jfirebaugh

PR #559, opened 2016-11-04
Don't rely on splitting on periods; explicitly disallow arbitrary keys in transitions

lbud:
Alternative to #558.

While it's doubtful anyone _is_ making up arbitrary keys in transitions in a style, I'd still rather not make the spec any _less_ strict — object validation now relies on a boolean option to error on arbitrary keys (so any other properties that go through vanilla object validation can have arbitrary keys).

@jfirebaugh @lucaswoj

### DIFF
--- a/lib/validate/validate_object.js
+++ b/lib/validate/validate_object.js
@@ -35,7 +35,7 @@ module.exports = function validateObject(options) {
             }));
 
         // tolerate root-level extra keys & arbitrary layer properties
-        } else if (key !== '' && options.disallowArbitrary) {
+        } else if (key !== '' && !options.allowArbitrary) {
             errors.push(new ValidationError(key, object[objectKey], 'unknown property "%s"', objectKey));
         }
     }
--- a/lib/validate/validate_object.js
+++ b/lib/validate/validate_object.js
@@ -2,13 +2,13 @@
 
 var ValidationError = require('../error/validation_error');
 var getType = require('../util/get_type');
-var validate = require('./validate');
+var validateSpec = require('./validate');
 
 module.exports = function validateObject(options) {
     var key = options.key;
     var object = options.value;
-    var valueSpec = options.valueSpec;
-    var objectElementValidators = options.objectElementValidators || {};
+    var elementSpecs = options.valueSpec || {};
+    var elementValidators = options.objectElementValidators || {};
     var style = options.style;
     var styleSpec = options.styleSpec;
     var errors = [];
@@ -19,30 +19,37 @@ module.exports = function validateObject(options) {
     }
 
     for (var objectKey in object) {
-        var valueSpecKey = objectKey.split('.')[0]; // treat 'paint.*' as 'paint'
-        var objectElementSpec = valueSpec && (valueSpec[valueSpecKey] || valueSpec['*']);
-        var objectElementValidator = objectElementValidators[valueSpecKey] || objectElementValidators['*'];
+        var elementSpecKey = objectKey.split('.')[0]; // treat 'paint.*' as 'paint'
+        var elementSpec = elementSpecs[elementSpecKey] || elementSpecs['*'];
 
-        if (objectElementSpec || objectElementValidator) {
-            errors = errors.concat((objectElementValidator || validate)({
-                key: (key ? key + '.' : key) + objectKey,
-                value: object[objectKey],
-                valueSpec: objectElementSpec,
-                style: style,
-                styleSpec: styleSpec,
-                object: object,
-                objectKey: objectKey
-            }));
-
-        // tolerate root-level extra keys & arbitrary layer properties
-        } else if (key !== '' && !options.allowArbitraryKeys) {
+        var validateElement;
+        if (elementValidators[elementSpecKey]) {
+            validateElement = elementValidators[elementSpecKey];
+        } else if (elementSpecs[elementSpecKey]) {
+            validateElement = validateSpec;
+        } else if (elementValidators['*']) {
+            validateElement = elementValidators['*'];
+        } else if (elementSpecs['*']) {
+            validateElement = validateSpec;
+        } else {
             errors.push(new ValidationError(key, object[objectKey], 'unknown property "%s"', objectKey));
+            continue;
         }
+
+        errors = errors.concat(validateElement({
+            key: (key ? key + '.' : key) + objectKey,
+            value: object[objectKey],
+            valueSpec: elementSpec,
+            style: style,
+            styleSpec: styleSpec,
+            object: object,
+            objectKey: objectKey
+        }));
     }
 
-    for (valueSpecKey in valueSpec) {
-        if (valueSpec[valueSpecKey].required && valueSpec[valueSpecKey]['default'] === undefined && object[valueSpecKey] === undefined) {
-            errors.push(new ValidationError(key, object, 'missing required property "%s"', valueSpecKey));
+    for (elementSpecKey in elementSpecs) {
+        if (elementSpecs[elementSpecKey].required && elementSpecs[elementSpecKey]['default'] === undefined && object[elementSpecKey] === undefined) {
+            errors.push(new ValidationError(key, object, 'missing required property "%s"', elementSpecKey));
         }
     }
 
--- a/lib/validate/validate_object.js
+++ b/lib/validate/validate_object.js
@@ -36,7 +36,7 @@ module.exports = function validateObject(options) {
 
         // tolerate root-level extra keys & arbitrary layer properties
         // TODO remove this layer-specific logic
-        } else if (key !== '' && key.split('.').length !== 1) {
+        } else if (key !== '' && options.disallowArbitrary) {
             errors.push(new ValidationError(key, object[objectKey], 'unknown property "%s"', objectKey));
         }
     }
--- a/lib/validate/validate_object.js
+++ b/lib/validate/validate_object.js
@@ -35,7 +35,6 @@ module.exports = function validateObject(options) {
             }));
 
         // tolerate root-level extra keys & arbitrary layer properties
-        // TODO remove this layer-specific logic
         } else if (key !== '' && options.disallowArbitrary) {
             errors.push(new ValidationError(key, object[objectKey], 'unknown property "%s"', objectKey));
         }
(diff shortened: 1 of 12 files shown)


lib/validate/validate_object.js
```diff
@@ -36,7 +36,7 @@ module.exports = function validateObject(options) {
 
         // tolerate root-level extra keys & arbitrary layer properties
         // TODO remove this layer-specific logic
```
lucaswoj: We can remove this todo now.

lucaswoj (changes requested):
This is good. Thank you for thinking through the implications of making the style spec stricter.

A couple questions:
- Should `allowArbitrary` be `true` by default? I would like to make new style spec primitives as strict as possible. Strictness should be the default.
- Are you sure that `transition` objects are the only ones that have been enforcing "strictness"? What about `light` objects? What about `layout` / `paint` objects within layers? What about function objects? 
- We already have a means of interacting with arbitrary keys in objects: the [`*` wildcard object element validator](http://github.com/mapbox/mapbox-gl-style-spec/blob/449f8431be19f242a1d9bbde3ac35c60d8997883/lib/validate/validate_object.js#L24-L25). What would it mean to specify a `*` wildcard object element validator and make `allowArbitrary` `false`? Could the two concepts be combined? For example, we could allow arbitrary keys iff a `*` wildcard object element validator is specified.
- Is `allowArbitrary` a specific enough name for this setting that someone who stumbles across it in the codebase without context would know what it means? "Allow arbitrary _what_?"

lbud:
@lucaswoj @jfirebaugh 

To make sure I fully understood this, I broke down the validation pipeline here (and caught a few bugs with the light validation I had written here :pray:):

Here's what happens when we validate a style:
- A full style enters the validation pipeline [here](https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/lib/validate_style.min.js#L27-L36). It (`$root`) is a non-function object, and it doesn't have a spec-defined type or a dedicated validator, so it [enters](https://github.com/mapbox/mapbox-gl-style-spec/blob/arbitrary-keys/lib/validate/validate.js#L62-L64) the object validator 
- The keys at the root level either go into their own object element validators, or they're arbitrary keys and at the root level they are tolerated.
  - `version`: enum validator
  - `name`: string validator
  - `metadata`: [wildcard passthrough validator](https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/lib/validate/validate.js#L21-L23)
  - `center`: array validator -> number validator
  - `zoom`: number validator
  - `bearing`: number validator
  - `pitch`: number validator
  - `light`: light validator
    - `*-transition`: [validated as `transition` types](https://github.com/mapbox/mapbox-gl-style-spec/blob/arbitrary-keys/lib/validate/validate_light.js#L42), `disallowArbitrary=true`
    - other keys: validated as one of (array, number, enum, color)
  - `sources`: source validator. Each key is validated through the `*` line [here](https://github.com/mapbox/mapbox-gl-style-spec/blob/arbitrary-keys/lib/validate/validate_object.js#L23)/[here](https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/reference/v8.json#L103) as a `source` type
    - `vector` and `raster` sources _with_ a `url` specified explicitly disallow arbitrary keys [here](https://github.com/mapbox/mapbox-gl-style-spec/blob/arbitrary-keys/lib/validate/validate_source.js#L30-L36)
    - none of the other types (or `vector` or `raster` sources without the `url` key) seem to explicitly disallow arbitrary keys — if this is the case it also seems like something we would break if we started enforcing `disallowArbitrary` only now
  - `sprite`: string validator
  - `glyphs`: [glyph validator](https://github.com/mapbox/mapbox-gl-style-spec/blob/arbitrary-keys/lib/validate/validate_glyphs_url.js)
  - `transition` ($root-level): **would validate as a vanilla object and thus allow arbitrary keys**. Should we add something to disallow these?
  - `layers`: array validator -> layer validator
    - ultimately these get to this [megavalidator](https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/lib/validate/validate_layer.js#L72-L107):
      - `filter`: [filter validator](https://github.com/mapbox/mapbox-gl-style-spec/blob/arbitrary-keys/lib/validate/validate_filter.js) (does not re-enter object validation)
      - `layout`: [re-enters the object validator with every key/value being subjected to the layout prop validator](https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/lib/validate/validate_layer.js#L88), effectively disallowing arbitrary sub-`layout` keys
        - individual layout properties: [layout prop validator](https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/lib/validate/validate_layout_property.js); [no layout properties are non-validator-primitive types](https://github.com/mapbox/mapbox-gl-style-spec/blob/ff8e72ae7702e68b2e69f6c54a38fa05ceac02e0/reference/v8.json#L352-L734) (i.e. re-enter object validator)
        - via the previous `.split('.')` mechanism arbitrary keys were disallowed
      - `paint`: [re-enters the object validator with every key/value being subjected to the paint prop validator](https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/lib/validate/validate_layer.js#L102), effectively disallowing arbitrary sub-`paint` keys
        - individual paint properties: [paint prop validator](https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/lib/validate/validate_paint_property.js); [no vanilla paint properties are non-validator-primitive types](https://github.com/mapbox/mapbox-gl-style-spec/blob/ff8e72ae7702e68b2e69f6c54a38fa05ceac02e0/reference/v8.json#L1578-L2699) (i.e. re-enter the object validator) except `-transition` properties
        - via the previous `.split('.')` mechanism arbitrary keys were disallowed
        - function values enter the function validator; by the previous `.split('.')` mechanism arbitrary keys were disallowed
      - `id`: string validator
      - `type`: enum validator
      - `metadata`: wildcard passthrough validator
      - `ref`: string validator (+ does additional ref validation [here](https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/lib/validate/validate_layer.js#L34-L53))
      - `source`: string validator
      - `source-layer`: string validator
      - `minzoom`: number validator
      - `maxzoom`: number validator
      - any other keys are permitted as arbitrary keys

---

> - Should `allowArbitrary` be `true` by default? I would like to make new style spec primitives as strict as possible. Strictness should be the default.

To clarify, it's `disallowArbitrary` in this PR but your point stands. We _could_ switch this to `allowArbitrary` (false by default) and allow it only in the root object and root layers.

> - Are you sure that `transition` objects are the only ones that have been enforcing "strictness"? What about `light` objects? What about `layout` / `paint` objects within layers? What about function objects?

See above. (`light` object has its own validator, which I just fixed some bugs in and added tests to make sure we don't tolerate arbitrary keys. (Note that the root-level `transition` _hasn't_ been prohibiting arbitrary keys — should it?) As shown in https://github.com/mapbox/mapbox-gl-style-spec/pull/562, layout and paint objects as well as function keys have been enforcing strictness via the splitting-on-periods mechanism. (I merged that PR then rebased here to ensure this enforces strictness in the same way.)

> - We already have a means of interacting with arbitrary keys in objects: the [`*` wildcard object element validator](http://github.com/mapbox/mapbox-gl-style-spec/blob/449f8431be19f242a1d9bbde3ac35c60d8997883/lib/validate/validate_object.js#L24-L25). What would it mean to specify a `*` wildcard object element validator and make `allowArbitrary` `false`? Could the two concepts be combined? For example, we could allow arbitrary keys iff a `*` wildcard object element validator is specified.

The wildcard object element validator is more for when `"type": "*"`, rather than for interacting with wildcard keys themselves. I'm not sure we should conflate these?

> - Is `allowArbitrary` a specific enough name for this setting that someone who stumbles across it in the codebase without context would know what it means? "Allow arbitrary _what_?"

I'm happy to change to `disallowArbitraryKeys` (or, if we go with the change mentioned above, `allowArbitraryKeys`).

---

In my opinion then the outstanding questions are
- should we add strictness (`disallowArbitrary`) to the `$root`-level global transition?
- ~~should we rename this from `disallowArbitrary` to `allowArbitrary`? (`.concat('keys'))~~ see below — done
- ~~should we toggle this to strict by default, and instead be explicit in all the places where we _do_ allow arbitrary keys, so that future object validation will be strict by default?~~ see below — done

lbud:
> - should we rename this from disallowArbitrary to allowArbitrary? (`.concat('keys'))
> - should we toggle this to strict by default, and instead be explicit in all the places where we do allow arbitrary keys, so that future object validation will be strict by default?

Just did in https://github.com/mapbox/mapbox-gl-style-spec/pull/559/commits/428bf80f1941ed3ebe76367db64f849c8a0f4075 https://github.com/mapbox/mapbox-gl-style-spec/pull/559/commits/f902b4a9a45aa679f068544f6834749e4db6d4f7

lucaswoj:
Thank you for being so thorough on this 🙇 

---

> should we add strictness (`disallowArbitrary`) to the `$root`-level global transition?

A stated goal of this PR is to not increase strictness. We should not add strictness to the `$root`-level global transition. 

> The wildcard object element validator is more for when `"type": "*"`, rather than for interacting with wildcard keys themselves. I'm not sure we should conflate these?

What would it mean to specify a `*` wildcard object element validator and when `allowArbitrary` false? Or is that combination impossible for a reason I'm missing? (It's been a while since I wrote this code.)

It is desirable that it be impossible to pass contradictary arguments to this method. It is desirable that objects which allow arbitrary keys must go against the grain to omit validation for arbitrary keys.

lib/validate/validate_object.js
```diff
@@ -35,8 +35,7 @@ module.exports = function validateObject(options) {
             }));
 
         // tolerate root-level extra keys & arbitrary layer properties
-        // TODO remove this layer-specific logic
-        } else if (key !== '' && key.split('.').length !== 1) {
+        } else if (key !== '' && !options.allowArbitraryKeys) {
```
jfirebaugh: Do we need the `key !== ''` part of this conditional? It looks like that's covered by passing `allowArbitraryKeys: true` in `validateStyleMin`.

jfirebaugh (approved):
Looks good to me other than the small question below. Thanks for the thoroughness of your investigation.

lucaswoj (approved):

PR #571, opened 2016-11-15
Remove/re-add layers when source-related properties change

### LINKED ISSUE #570: Handle changes to a layer's 'source', 'source-layer', or 'type' in diffStyles
Right now, changes to a layer's `source`, source-layer`, or `type` produce a corresponding `setLayerProperty` operation from diffStyles.  On the GL JS side, I am pretty sure that the way to handle this will be to remove the layer and create/add a new one.  

Does it make more sense for this to happen here or downstream where the change operations are being applied?  

Doing it here bakes in the assumption that these properties of a style layer are immutable; doing it downstream means there might be some unweildy logic to watch for cases where one of these operations happens in conjunction with other changes to the layer.

anandthakker:
Two cases:

- Fix #570 - remove/re-add layer when `source`, `source-layer`, or
  `type` properties are changed.
- Remove all dependent layers before removing a source; if the source is
  being removed and re-added, then the corresponding layers are also
  added again afterward. (See https://github.com/mapbox/mapbox-gl-js/pull/3621#issuecomment-260786790)

### DIFF
--- a/lib/diff.js
+++ b/lib/diff.js
@@ -97,7 +97,7 @@ var operations = {
 };
 
 
-function diffSources(before, after, commands) {
+function diffSources(before, after, commands, sourcesRemoved) {
     before = before || {};
     after = after || {};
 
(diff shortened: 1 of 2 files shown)


jfirebaugh (approved):

anandthakker:
@scothis @tmcw Heads up that this just merged -- some changes to the diff module.  Hopefully shouldn't hurt anything, studio-wise, but just in case

PR #573, opened 2016-11-16
Add declass module to preprocess a style's paint classes

anandthakker:


anandthakker:
Prepares for removal of paint classes from GL JS by providing `declass` as a migration path for projects relying on them.  `map.setClasses(classes)` will become `map.setStyle(declass(myBaseStyleJSON, classes))`.

Couple notes:
- Currently I have `declass` leaving the `paint.foo` properties on layers, even after applying the `foo` class, but I'm not sure whether that's a good idea.
- Based on my admittedly brief skim of GL JS paint class code, I believe there's no provision for a paint class to _unset_ a property, and so that's the case here as well.  One implication of this is that users may need to hang on to their original style JSON if they want to flip back and forth between style-with-no-class-applied and style-with-a-class-applied -- or else they'll need to explicitly declare in the base `paint` definition the value for any property that's defined in a `paint.blah` definition

jfirebaugh:
> - Currently I have `declass` leaving the `paint.foo` properties on layers, even after applying the `foo` class, but I'm not sure whether that's a good idea.

It should remove them -- the output should include only plain `paint` properties.

> - Based on my admittedly brief skim of GL JS paint class code, I believe there's no provision for a paint class to _unset_ a property, and so that's the case here as well.

Correct, they are additive only.

anandthakker:
> It should remove them -- the output should include only plain paint properties

@jfirebaugh removed.

lib/declass.js
```diff
@@ -0,0 +1,42 @@
...
+    });
+};
+
+function declassLayer(layer, klass) {
+    return extend(layer, {
+        paint: extend(layer.paint, layer['paint.' + klass])
+    });
+}
+
+function extend(dest, source) {
```
jfirebaugh: Use lib/util/extend instead.

jfirebaugh (changes requested):

lib/declass.js
anandthakker: Whoops, didn't see that.  Switched.

jfirebaugh (approved):

PR #576, opened 2016-11-17
Drop paint classes

anandthakker:
- Add deprecation notice for class-specific paint properties to `v8.json` and narrative docs.
 - Update API docs with declassStyle so that it can be referenced in ☝️
 - Create `v9.json`, and remove `paint.*`.
 - Initialize `migrations/v9` with deref & declass functionality.


anandthakker:
Should `v9.json` be renamed to `next.json` (or something) so as not to imply that v9 is finished?

jfirebaugh:
Let's not create `v9.json` yet. Adding it now means we have to keep it in sync with changes in `v8.json`.

anandthakker:
👍  sounds good.  Shall we keep `migrations/v9`, since that won't have to change unless there's a new migration?

docs/_generate/index.html
```diff
@@ -628,9 +628,8 @@ <h3 class='space-bottom1'><a href='#sources-video' title='link to video'>video</
       <p>
         <em>Paint properties</em> are applied later in the rendering process. A layer that shares layout properties with another
         layer can have independent paint properties. Paint properties appear in the layer's <code>"paint"</code> object.
-        Layers can also have class-specific paint properties, which are applied only when the map has a certain class
-        name set. For example, a layer with a <code>"paint.night"</code> property would have those properties applied
-        when the map has the <code>"night"</code> class set.
+        Layers can also have class-specific paint properties, defined with a property like <code>"paint.night"</code> and applied only when the map has a certain class
```
jfirebaugh: Let's just remove this.

API.md
```diff
@@ -60,6 +107,29 @@ fs.writeFileSync('./dest.min.json', format(style, 0));
 **Returns** `string`, stringified formatted JSON
 
 
+### `exports(layers)`
```
jfirebaugh: Can you fix the naming like you did with `derefLayers`? Should be `groupByLayout`.

migrations/v9.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+var deref = require('../lib/deref');
+
+function eachLayer(style, callback) {
+    for (var k in style.layers) {
+        callback(style.layers[k]);
+        eachLayer(style.layers[k], callback);
```
jfirebaugh: This shouldn't be recursive.

jfirebaugh (changes requested):

anandthakker:
@jfirebaugh ^ changes made.  Back to you.

jfirebaugh (approved):

PR #579, opened 2016-11-22
Add support for circle-stroke-* properties

lucaswoj:


reference/v8.json
```diff
@@ -2213,6 +2213,61 @@
...
+          "js": "master"
+        },
+        "data-driven styling": {
+          "js": "master"
+        }
+      }
+    },
+    "circle-stroke-opacity": {
+      "type": "number",
+      "doc": "The stroke opacity of the circle.",
```
lbud: This sounds a little funky to me — in my head something like "The opacity of the circle's stroke" might sound better? (but if this sounds fine to other people I am okay with it)

lucaswoj: Agreed. Updated in the latest commit. (Along with some other tweaks).

lbud (approved):

PR #580, opened 2016-11-22
Define a clear order of operations for object element validation, remove "allowArbitraryKeys" option

lucaswoj:
This PR refactors `validateObject` such that

 - the order of precedence for object element validation logic is more intuitive and useful:
   1. a specific object element validator
   2. a specific object element specification
   3. a wildcard object element validator
   4. a wildcard object element specification
   5. error!
 - the `allowArbitraryKeys` option is rolled into the above logic (case "v" ☝️)
 - internal variable names are more consistent, giving better contextual clues to readers about how the function works

cc @lbud @jfirebaugh

### DIFF
--- a/lib/validate/validate_layer.js
+++ b/lib/validate/validate_layer.js
@@ -74,8 +74,10 @@ module.exports = function validateLayer(options) {
         valueSpec: styleSpec.layer,
         style: options.style,
         styleSpec: options.styleSpec,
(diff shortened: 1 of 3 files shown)


lucaswoj:
ref #559

lbud:
@lucaswoj this looks good to me. (Should this just be re-targeted at mb-pages and we can close #559?)

PR #591, opened 2016-11-29
"visibility" property does not support functions

### LINKED ISSUE #442: visibility properties can't be functions
The documentation advertises that the `visibility` layout property for any layer type can be a function. As far as I can tell, this isn't true: neither gl-js nor gl-native nor Mapbox Studio support using functions for `visibility`. Furthermore, the use case for doing so seems marginal:
- Using a non-data-driven function for `visibility` would be basically equivalent to using `minzoom` and `maxzoom`. A `visibility` function is slightly more flexible; for example you could have a layer be _hidden_ between a certain minimum and maximum zoom, but visible outside of it.
- Using a data-driven function for `visibility` would provide a subset of what `filter` can do.

I think we should revise the documentation to indicate that `visibility` cannot be a function.

jfirebaugh:
Fixes #442.

This reflects the de-facto situation in Mapbox GL JS and Native: a `visibility` property value cannot be a function.

This will _increase_ the strictness of the validator, possibly causing some existing styles to fail validation where they would previously pass. But the potential fallout in this case seems limited, since Studio has never allowed creating such a style, and it wouldn't work as anticipated in our SDKs anyway.

cc @lucaswoj @mapbox/appbox


scothis:
We've been burned by adding validation in the past without also migrating styles, but this is one that seems safe.

mollymerp (approved):
LGTM 🚢 on 💚

Issue #595, opened 2016-11-29
Why are stop values for property functions allowed to be an array?


jfirebaugh:
Why is array included in [this list](https://github.com/mapbox/mapbox-gl-style-spec/blob/d91d3cd4bce435a75cebf8fd17501bfadfa046e4/lib/validate/validate_function.js#L136-L138)? Do we really want to allow, e.g.:

```
{
  "type": "categorical",
  "property": "p",
  "stops": [
    [[0, 1], 0],
    [[2, 3], 10]
  ]
}
```

This seems like a crazy edge case that we shouldn't support.

cc @lucaswoj @ansis

ansis:
I agree it shouldn't be supported

lucaswoj:
This was a mistake on my part. The intention was to allow stops for range values, not domain values.

Issue #593, opened 2016-11-29
Prohibit mixed stop value types for zoom-and-property functions


jfirebaugh:
The validator prohibits:

```
{
  "property": "p",
  "stops": [[0, 0], ["foo", 10]]
}
```

It should also prohibit:

```
{
  "property": "p",
  "stops": [[{zoom: 0, value: 0}, 0], [{zoom: 0, value: "foo"}, 10]]
}
```

Issue #587, opened 2016-11-29
Non-categorical functions with string stops?


jfirebaugh:
How does this function work? E.g. what is the evaluated value for a feature whose `p` property is "fuzz"?

```
{
  "type": "exponential",
  "property": "p",
  "stops": [
    ["foo", 0],
    ["bar", 100]
  ]
}
```

What about this one? Is it valid?

```
{
  "type": "interval",
  "property": "p",
  "stops": [
    ["a", 0],
    ["å", 5],
    ["b", 10]
  ]
}
```

cc @lucaswoj

lucaswoj:
String domains only make sense for `categorical` functions. `exponential` and `interval` functions require a numeric domain.

PR #596, opened 2016-11-29
Fix function validation issues

jfirebaugh:
Fixes for #587, #593, and #595.

cc @lucaswoj @ansis

### DIFF
--- a/lib/validate/validate_function.js
+++ b/lib/validate/validate_function.js
@@ -13,8 +13,8 @@ module.exports = function validateFunction(options) {
     var functionValue = options.value;
     var stopKeyType;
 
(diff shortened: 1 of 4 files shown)


lib/validate/validate_function.js
```diff
@@ -143,8 +143,8 @@ module.exports = function validateFunction(options) {
             return [new ValidationError(options.key, options.value, '%s stop domain type must match previous stop domain type %s', type, stopKeyType)];
```
lucaswoj, line 143: Do you think we should rename this function to `validateDomainValue` to ensure mistakes like this are more obvious in the future?

jfirebaugh: Renamed.

lucaswoj (approved):

PR #605, opened 2016-11-30
Add sdk-support for light properties

scothis:
Only supported by GL JS 0.27+


scothis:
cc @lbud

jfirebaugh:
These do show up in docs automatically:

image

jfirebaugh (approved):

PR #608, opened 2016-11-30
Function validation

jfirebaugh:
Followup to #596, next round of fixes.

cc @lucaswoj @davidtheclark


lucaswoj (approved):
Looks good! Thanks @jfirebaugh. 🙇

PR #615, opened 2016-12-07
Add support for icon-offset property functions

jellevoost:


lucaswoj:
A quick note to trigger Github's crosslinking functionality: this is related to https://github.com/mapbox/mapbox-gl-js/pull/3729

jfirebaugh:
`icon-offset` units are relative to `icon-size` (#84). Do we face a similar issue with making `icon-offset` data-driven as we did with `line-width` and `line-dasharray`?

cc @ansis

ansis:
> icon-offset units are relative to icon-size (#84). Do we face a similar issue with making icon-offset data-driven as we did with line-width and line-dasharray?

Nope, because they are relative to the `icon-size` at the current zoom level.

Side note: `icon-offset` was made relative to `icon-size` in order to make it work more easily with the collision detection, not as what-is-better-externally decision.

jfirebaugh:
Committed as f7d472426f0235ed3fe1e4721b5d68f3c171010a. Thanks!

PR #616, opened 2016-12-08
Remove "property-function" on properties whose DDS status is undecided

### LINKED ISSUE #592: "property-function" should be false until supported in an SDK
Until we know for sure that we'll support data-driven styling for a particular property, the value for `"property-function"` should be `false`, for validation and Studio's purposes.

* It's easy to go from "property function not allowed" to "property function allowed", but going the other direction risks causing validation problems for existing styles.
* Studio should be able to use the value of this property to decide whether to offer data-driven styling for a particular property. Right now the values aren't useful, because they've all been defaulted to `true`.

cc @lucaswoj

jfirebaugh:
* `"property-function": true` means DDS is enabled and implemented in at least one SDK
* `"property-function": false` means we've decided this property will not support DDS
* Lack of a value means we haven't decided, or haven't implemented it yet

Fixes #592 

cc @lucaswoj @davidtheclark


test/fixture/functions.output.json
```diff
@@ -52,22 +52,42 @@
...
   {
+    "message": "layers[13].paint.fill-antialias: property functions not supported",
+    "line": 241
+  },
+  {
     "message": "layers[13].paint.fill-antialias.stops[0]: object stop key must have zoom",
     "line": 244
   },
   {
+    "message": "layers[14].paint.fill-antialias: property functions not supported",
```
lucaswoj: It's probably unwise to have our style validator tests for property functions be based upon an unsupported property (`fill-antialias`). Could you switch these to another property?

reference/v8.json
```diff
@@ -2605,7 +2540,6 @@
       "type": "enum",
       "function": "piecewise-constant",
       "zoom-function": true,
-      "property-function": true,
       "values": {
         "map": {
             "doc": "The text is translated relative to the map."
```
lucaswoj, line 2545: Went through ☝️ manually and everything looks 👌

test/fixture/functions.output.json
jfirebaugh: Yep, fixed.

lucaswoj (approved):
🐑 it!

PR #617, opened 2016-12-08
Eliminate fill-extrusion-base's special case "requires" value

jfirebaugh:
The code generator in mapbox-gl-native doesn't know how to deal with the `"<="` key. Since no other `requires` values use comparison operators, and this comparison can't in most cases be enforced because the actual values will come from source data, I think it's better to note this restriction in prose documentation instead.


lbud (approved):
👍

PR #631, opened 2016-12-22
Move package to @mapbox namespace.

tmcw:


lucaswoj:
Thanks @tmcw & @tmcw's bash scripts. 

I think this is the workflow we'll need to get this change pushed all the way through. Anything missing here? 

# Todo

 - [ ] 🚢
 - [ ] publish this package in @mapbox namespace
 - [ ] deprecate the old non-namespaced package
 - [ ] update all internal projects that reference `mapbox-gl-style-spec`

mourner:
How about removing the `mapbox-` prefix for less redundancy? E.g. `@mapbox/gl-style-spec` instead of `@mapbox/mapbox-gl-style-spec`.

jfirebaugh:
If we do that, we should remove `gl` too.

mourner:
@jfirebaugh should we also publish GL JS under `@mapbox/js`, or should it still be `@mapbox/gl-js`?

lucaswoj:
Planning to do this at the same time as https://github.com/mapbox/mapbox-gl-js/issues/3794

lucaswoj:
Rolling this into mapbox/mapbox-gl-js#3794 and https://github.com/mapbox/mapbox-gl-js/pull/3946

PR #641, opened 2016-12-28
Indicate macOS SDK support

1ec5:
Added macOS SDK support information alongside the existing iOS SDK support information. (macOS SDK v0.2.1 [supports a few properties](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/macos/CHANGELOG.md#021---july-19-2016) that iOS SDK v3.3.x does not.) Also clarified that, with respect to the iOS and macOS SDKs, this document is intended as a file format reference for style authors rather than a runtime styling reference.

/cc @jfirebaugh

### DIFF
--- a/docs/_generate/index.html
+++ b/docs/_generate/index.html
@@ -105,10 +105,11 @@ navigation:
       <p>The intended audience of this specification includes:
       <ul>
       <li>Advanced designers and cartographers who want to write styles by hand rather than use <a href='https://www.mapbox.com/studio'>Mapbox Studio</a></li>
-      <li>Developers using style-related features of <a href='https://www.mapbox.com/mapbox-gl-js/'>Mapbox GL JS</a> or the Mapbox <a href='https://www.mapbox.com/ios-sdk/'>iOS</a> or <a href='https://www.mapbox.com/android-sdk/'>Android</a> SDKs</li>
+      <li>Developers using style-related features of <a href='https://www.mapbox.com/mapbox-gl-js/'>Mapbox GL JS</a> or the <a href='https://www.mapbox.com/android-sdk/'>Mapbox Android SDK</a></li>
       <li>Authors of software that generates or processes Mapbox styles.</li>
       </ul>
       </p>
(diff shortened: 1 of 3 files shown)


docs/_generate/index.html
```diff
@@ -105,10 +105,11 @@
       <p>The intended audience of this specification includes:
       <ul>
       <li>Advanced designers and cartographers who want to write styles by hand rather than use <a href='https://www.mapbox.com/studio'>Mapbox Studio</a></li>
-      <li>Developers using style-related features of <a href='https://www.mapbox.com/mapbox-gl-js/'>Mapbox GL JS</a> or the Mapbox <a href='https://www.mapbox.com/ios-sdk/'>iOS</a> or <a href='https://www.mapbox.com/android-sdk/'>Android</a> SDKs</li>
+      <li>Developers using style-related features of <a href='https://www.mapbox.com/mapbox-gl-js/'>Mapbox GL JS</a> or the <a href='https://www.mapbox.com/android-sdk/'>Mapbox Android SDK</a></li>
       <li>Authors of software that generates or processes Mapbox styles.</li>
       </ul>
       </p>
+      <p>Developers using the <a href='https://www.mapbox.com/ios-sdk/'>Mapbox iOS SDK</a> or <a href='https://github.com/mapbox/mapbox-gl-native/tree/master/platform/macos/'>Mapbox macOS SDK</a> should consult the <a href="https://www.mapbox.com/ios-sdk/api/3.4.0-beta.6/">iOS SDK API reference</a> for platform-appropriate documentation of style-related features.</p>
```
jfirebaugh: Can we use a link for the iOS SDK API that redirects to the latest documentation, rather than one which will need to be updated whenever we release a new version?

1ec5: This is a stopgap until we release v3.4.0 officially. At that point, removing the version from the URL will cause the link to always point to the latest documentation. I can leave out this link for now, though.

jfirebaugh: 👍  -- whatever's easier and more likely to be remembered when the time comes.

jfirebaugh (approved):

Issue #659, opened 2017-01-21
Canvas source type is undocumented


1ec5:
#656 added a `canvas` source type along with some documentation in JSON format. But it looks like the more substantial HTML documentation for sources is hard-coded in docs/_generate/index.html, which needs to be updated.

/cc @lbud @lucaswoj

PR #661, opened 2017-01-24
Add source-canvas documentation

lbud:
Closes https://github.com/mapbox/mapbox-gl-style-spec/issues/659.

image

### DIFF
--- a/docs/_generate/index.html
+++ b/docs/_generate/index.html
@@ -543,9 +543,9 @@ navigation:
             <tr>
               <td>basic functionality</td>
               <td>&gt;= 0.32.0</td>
-              <td><a href="https://github.com/mapbox/mapbox-gl-native/issues/7823">Not yet supported</a></td>
-              <td><a href="https://github.com/mapbox/mapbox-gl-native/issues/7823">Not yet supported</a></td>
-              <td><a href="https://github.com/mapbox/mapbox-gl-native/issues/7823">Not yet supported</a></td>
+              <td>Not supported</td>
+              <td>Not supported</td>
+              <td>Not supported</td>
             </tr>
             </tbody>
           </table>
--- a/docs/_generate/index.html
+++ b/docs/_generate/index.html
@@ -34,6 +34,7 @@ navigation:
     - title: GeoJSON
     - title: Image
     - title: Video
+    - title: Canvas
   - title: Sprite
   - title: Glyphs
   - title: Transition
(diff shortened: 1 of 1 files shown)


docs/_generate/index.html
```diff
@@ -488,6 +489,67 @@ <h3 class='space-bottom1'><a href='#sources-video' title='link to video'>video</
             </tbody>
           </table>
         </div>
+        <div id='sources-canvas' class='pad2 keyline-bottom'>
+          <h3 class='space-bottom1'><a href='#sources-canvas' title='link to canvas'>canvas</a></h3>
+          <p>
+            A canvas source. The <code>"canvas"</code> value is the ID of the canvas element in the document.
```
1ec5, line 495: This is already in a `<code>` tag, so it doesn't need quotation marks. (×2)

docs/_generate/index.html
```diff
@@ -488,6 +489,67 @@ <h3 class='space-bottom1'><a href='#sources-video' title='link to video'>video</
...
+              <td class='center'>Android SDK</td>
+              <td class='center'>iOS SDK</td>
+              <td class='center'>macOS SDK</td>
+            </tr>
+            </thead>
+            <tbody>
+            <tr>
+              <td>basic functionality</td>
+              <td>&gt;= 0.32.0</td>
+              <td><a href="https://github.com/mapbox/mapbox-gl-native/issues/7823">Not yet supported</a></td>
```
1ec5: mapbox/mapbox-gl-native#7823 only applies to the iOS and macOS SDKs; I have no idea whether it makes sense on Android. Moreover, it's unlikely that any gl-native feature will look anything like this canvas source type. "Not supported" would be more accurate.

1ec5: "Not supported" should be unlinked as well.

1ec5 (changes requested):

docs/_generate/index.html
lbud: Hmm…much of this documentation _does_ contain `<code>"something"</code>` — it seems to generally apply to many JSON values https://www.mapbox.com/mapbox-gl-style-spec/#sources. I copied this style from video sources, right above it ([e.g.](https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/docs/_generate/index.html#L439-L444))

1ec5: Ah, I guess this is an inconsistency between the source and layer documentation. Fodder for other PRs, I guess.

1ec5 (approved):

PR #666, opened 2017-01-26
Remove support for implicitly-categorical functions

jfirebaugh:
This turned out to add significant complexity to the gl-native implementation, and never took effect in gl-js due to a bug (https://github.com/mapbox/mapbox-gl-js/issues/3717). Removing it due to the complexity it would add to gl-native and other implementations, and the following additional rationales:

* As a general principle: explicit is better than implicit.
* Most users will not be constructing functions by hand anyway. For those that do:
* Requiring `"type": "categorical"` is not an undue burden, especially when coupled with a hinting mechanism in the validator as implemented here.
* Implicitly supplying `"type": "categorical"` for string domain values introduces a potentially surprising difference between `"stops": [["1", 1], ["2", 2]]` and `"stops": [[1, 1], [2, 2]]`.

Closes https://github.com/mapbox/mapbox-gl-js/issues/3717
Closes https://github.com/mapbox/mapbox-gl-js/issues/3948

### DIFF
--- a/lib/validate/validate_function.js
+++ b/lib/validate/validate_function.js
@@ -149,9 +149,6 @@ module.exports = function validateFunction(options) {
 
         if (!stopKeyType) {
             stopKeyType = type;
-            if (!functionType && type === 'string') {
-                functionType = 'categorical';
-            }
         } else if (type !== stopKeyType) {
             return [new ValidationError(options.key, options.value, '%s stop domain type must match previous stop domain type %s', type, stopKeyType)];
         }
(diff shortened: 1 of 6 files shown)


lib/validate/validate_function.js
```diff
@@ -161,7 +158,11 @@ module.exports = function validateFunction(options) {
         }
 
         if (type !== 'number' && functionType !== 'categorical') {
-            return [new ValidationError(options.key, options.value, 'number expected, %s found', type)];
+            var message = 'number expected, %s found';
+            if (functionValueSpec['property-function'] && functionType === undefined) {
+                message += '\nIf you intended to use a categorical function, specify `"type": "categorical"`.';
```
lucaswoj, line 163: This will help so much 👌

lucaswoj (approved):

PR #667, opened 2017-01-26
Add a validation hint for mistaken attempts to use interpolation syntax

### LINKED ISSUE #664: Validator should check for attempts to use {interpolation} syntax in DDS properties
Style validation should check for attempts to use "{interpolation}" syntax in properties that aren't `text-field` or `icon-image`, and output an error message that provides the correct use of a data-driven property.

E.g. if the validator sees `"icon-rotate": "{rotation}"`, the error message could be: "The `icon-rotate` property does not support string interpolation syntax. Use a property function instead: `"icon-rotate": { "type": "identity", "property": "rotation" }`.

cc @geografa

jfirebaugh:
Fixes #664

### DIFF
--- a/lib/validate/validate_property.js
+++ b/lib/validate/validate_property.js
@@ -2,6 +2,7 @@
 
 var validate = require('./validate');
 var ValidationError = require('../error/validation_error');
+var getType = require('../util/get_type');
 
 module.exports = function validateProperty(options, propertyType) {
     var key = options.key;
@@ -27,6 +28,13 @@ module.exports = function validateProperty(options, propertyType) {
     var valueSpec = options.valueSpec || layerSpec[propertyKey];
     if (!valueSpec) {
         return [new ValidationError(key, value, 'unknown property "%s"', propertyKey)];
+    }
+
+    var tokenMatch;
+    if (getType(value) === 'string' && valueSpec['property-function'] && !valueSpec.tokens && (tokenMatch = /^{([^}]+)}$/.exec(value))) {
+        return [new ValidationError(key, value, '"%s" does not support interpolation syntax\n' +
+            'Use an identity property function instead: `{ "type": "identity", "property": %s` }`.',
+            propertyKey, JSON.stringify(tokenMatch[1]))];
     }
 
     var errors = [];
--- a/lib/validate/validate_property.js
+++ b/lib/validate/validate_property.js
@@ -1,0 +1,49 @@
+'use strict';
+
+var validate = require('./validate');
+var ValidationError = require('../error/validation_error');
+
+module.exports = function validateProperty(options, propertyType) {
+    var key = options.key;
+    var style = options.style;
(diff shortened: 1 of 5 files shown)


lib/validate/validate_property.js
```diff
@@ -0,0 +1,57 @@
...
+    var valueSpec = options.valueSpec || layerSpec[propertyKey];
+    if (!valueSpec) {
+        return [new ValidationError(key, value, 'unknown property "%s"', propertyKey)];
+    }
+
+    var tokenMatch;
+    if (getType(value) === 'string' && valueSpec['property-function'] && !valueSpec.tokens && (tokenMatch = /^{([^}]+)}$/.exec(value))) {
+        return [new ValidationError(key, value, '"%s" does not support interpolation syntax\n' +
+            'Use an identity property function instead: `{ "type": "identity", "property": %s` }`.',
+            propertyKey, JSON.stringify(tokenMatch[1]))];
```
anandthakker, line 37: Why return early here with this error, rather than accumulating other possible errors below (i.e. via `validate(...)`)?

lib/validate/validate_property.js
```diff
@@ -0,0 +1,57 @@
...
+var ValidationError = require('../error/validation_error');
+var getType = require('../util/get_type');
+
+module.exports = function validateProperty(options, propertyType) {
+    var key = options.key;
+    var style = options.style;
+    var styleSpec = options.styleSpec;
+    var value = options.value;
+    var propertyKey = options.objectKey;
+    var layerSpec = styleSpec[propertyType + '_' + options.layerType];
```
anandthakker, line 13: ^ I think this will get tagged by the linter when we merge into the gl-js repo

anandthakker (commented):
👍 to the deduping and the nicely helpful suggestion in the new error message.

One question inline

lib/validate/validate_property.js
jfirebaugh: The follow-on error (usually "expected ___, got string") doesn't add much additional value and could be confusing.

lib/validate/validate_property.js
jfirebaugh: What about it specifically?

anandthakker: String concatenation with `+` instead of template literal

lib/validate/validate_property.js
anandthakker: 👍 cool, makes sense to me

anandthakker (approved):
Marking approved, pending the lint fix.

lib/validate/validate_property.js
jfirebaugh: Gotcha. So far we've avoided using template literals and other ES6 features in mapbox-gl-style-spec. If we were to start, we'd need to think about the downstream implications. (This applies to the monorepo too. cc @lucaswoj)

PR #668, opened 2017-01-26
Restore "data-driven styling" row for all style properties

### LINKED ISSUE #529: Restore "data-driven styling" row for all style properties
So that there is no ambiguity, it's useful to explicitly indicate which properties _don't_ support DDS yet, in any SDK. We used to have a "data-driven styling" row for all style properties, with "Not yet supported" in all columns, but it got lost in #501. Let's restore it.

jfirebaugh:
Fixes #529

### DIFF
--- a/reference/v8.json
+++ b/reference/v8.json
@@ -550,7 +550,8 @@
           "android": "2.0.1",
           "ios": "2.0.0",
           "macos": "0.1.0"
-        }
+        },
+        "data-driven styling": {}
       }
     },
     "line-join": {
@@ -576,7 +577,8 @@
           "android": "2.0.1",
           "ios": "2.0.0",
           "macos": "0.1.0"
-        }
+        },
+        "data-driven styling": {}
       }
     },
     "line-miter-limit": {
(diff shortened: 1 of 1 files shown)


reference/v8.json
```diff
@@ -1861,6 +1904,9 @@
       "sdk-support": {
         "basic functionality": {
           "js": "0.27.0"
+        },
+        "data-driven styling": {
+          "js": "Not supported"
```
anandthakker: I'm seeing this being treated like a version number in the generated docs:
screen shot 2017-01-26 at 7 09 21 pm

I think just removing the entry fixes it -- @jfirebaugh is there some other reason we need it in here?  If so we can tweak the html template to handle this...

anandthakker (changes requested):

reference/v8.json
jfirebaugh: Ah, whoops. Yeah, `fill-extrusion-opacity` is probably never going to support data-driven styling, for the reasons outlined in #545. I'll see if I can tweak the template to make this work.

jfirebaugh: Settling for a prose solution:

image

anandthakker (approved):
👍 LGTM

